PR #91, opened 2023-06-09
Use initContainers for django migrations

dsavineau:
The current django migration workflow is too complicated and splitted between:
- k8s batch/job (first run)
- k8s exec on API pod

Instead of doing this logic, we can just leverage the initContainers in the API deployment for running the django migration scripts (and for the initial data creation as well).

Currently the migration subcommand is executed on all operator runs but this isn't needed.
The migration subcommand should only has to be executed when the EDA container image changes so the deployment will be updated (and the init container triggered again).


rooftopcellist:
After talking with @hyagi, I think we could move back to the k8s job if desired and just run it on these conditions:

* if the image status is not set (status.image and status.version are set) - or if the api pod doesn't exist
* if the application api container image has changed

> if the k8s migration job already exists and is still running, don't create a new one (avoids conflicts)

PR #97, opened 2023-06-19
eda: Use gunicorn for running the API

dsavineau:
The django runserver command isn't designed to be ran in production but only for development workflow.
Instead we should use gunicorn since that's the workflow used for VM based installer.

### DIFF
--- a/config/crd/bases/eda.ansible.com_edas.yaml
+++ b/config/crd/bases/eda.ansible.com_edas.yaml
@@ -90,6 +90,14 @@ spec:
                   replicas: 1
                 description: Defines desired state of eda-api resources
                 properties:
+                  gunicorn_workers:
+                    default: 2
+                    description: 'The number of gunicorn workers for the api.
+                      Default: 2'
+                    format: int32
+                    minimum: 1
+                    nullable: false
+                    type: integer
                   node_selector:
                     additionalProperties:
                       type: string
--- a/config/manifests/bases/eda-server-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/eda-server-operator.clusterserviceversion.yaml
@@ -323,6 +323,12 @@ spec:
         path: api
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:advanced
+      - description: The number of gunicorn workers for the API.
+        displayName: Gunicorn API workers
+        path: api.gunicorn_workers
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:advanced
+        - urn:alm:descriptor:com.tectonic.ui:number
       - displayName: API server resource requirements
         path: api.resource_requirements
         x-descriptors:
--- a/roles/eda/defaults/main.yml
+++ b/roles/eda/defaults/main.yml
@@ -28,6 +28,7 @@ websocket_ssl_verify: false
 
 api: {}
 _api:
+  gunicorn_workers: 2
   replicas: 1
   resource_requirements:
     requests:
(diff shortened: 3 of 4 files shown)


rooftopcellist:
Currently, the upstream eda-server image does not include the gunicorn dependency, so when we deploy with this code, I see this error in the API container logs:

```
$ oc logs eda-demo-api-89bd897cc-ddwfj
Defaulted container "eda-api" out of: eda-api, daphne, eda-migrations (init), eda-initial-data (init)
/bin/bash: line 1: gunicorn: command not found
```

rooftopcellist:
The following PR to add gunicorn in the eda-server container is needed for this PR:
* https://github.com/ansible/aap-eda/pull/306

rooftopcellist:
After adding the dependency, the API deployment still doesn't fully come up, the livelinees probe is timing out, so the deployment keeps getting restarted:

```
  Warning  Unhealthy       44s (x2 over 84s)   kubelet            Liveness probe failed: Get "http://10.130.2.112:8000/_healthz": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

Increasing the timeouts for the readiness and liveliness probes fixes it.  I am trying to find the right balance now.  pulp-operator uses gunicorn for the api deployment and has the following probe settings:

* https://github.com/pulp/pulp-operator/blob/ansible/roles/pulp-api/templates/pulp-api.deployment.yaml.j2#L155-L172

That might be a good starting point

These settings worked for my deployment:
* https://github.com/rooftopcellist/aap-eda-operator/compare/increase-timeouts?expand=1

cutwater (changes requested):
@rooftopcellist While I agree with initial statement and encourage stop using `runserver` in production, simply running `gunicorn` is not sufficient as an alternative solution.

At the moment we serve both REST API and websockets via the `api` service. While REST API can be simply served via WSGI protocol, websockets require ASGI server (that can be handled either by [daphne](https://github.com/django/daphne) or [uvicorn](https://www.uvicorn.org/)).

There are two possible options available:

1. Serve both REST API and websockets by ASGI server (either daphne or uvicorn). In this case gunicorn is not needed.
NOTE: In some deployments gunicorn may together with uvicorn worker and act as a process manager and load balancer between uvicorn workers, but this is usually not needed in containerized environments (including K8S), since the container \ pod acts as a unit of scaling.

2. Serve REST API by WSGI server (gunicorn or uwsgi) and websockets by ASGI server. This will require spawning additional service and routing the traffic on a reverse proxy side.

For the sake of deployment simplicity I'd recommend the first option with performance testing afterwards that may change this decision in favor different option.

Alex-Izquierdo:
We already spawn dapnhe for asgi, so I think it's just a matter to replace runserver with gunicorn. https://github.com/ansible/eda-server-operator/blob/main/roles/eda/templates/eda-api.deployment.yaml.j2#L221

cutwater:
My apologies, I didn't see that daphne is already running. In this case this change makes perfect sense.

cutwater (approved):

rooftopcellist:
This PR worked with https://github.com/ansible/aap-eda/pull/306

But when testing with gunicorn installed via the Dockerfile, it doesn't work.  ([This merged PR](https://github.com/ansible/aap-eda/pull/307))

```
Defaulted container "eda-api" out of: eda-api, daphne, eda-migrations (init), eda-initial-data (init)
Traceback (most recent call last):
  File "/app/.local/bin/gunicorn", line 5, in <module>
    from gunicorn.app.wsgiapp import run
ModuleNotFoundError: No module named 'gunicorn'
```

Alex-Izquierdo:
I would say that it has to be installed in virtualenv

rooftopcellist (approved):
This works with eda-server images built from the most recent PR:
* https://github.com/ansible/aap-eda/pull/308

PR #104, opened 2023-07-10
Add scheduler deployment to handle more rulebook activations [DNM]

rooftopcellist:
This is the operator component to these changes, which add a standalone RQ scheduler.

*  ~https://github.com/ansible/aap-eda/pull/263~ (closed in favor of 321, which has merged)

* https://github.com/ansible/eda-server/pull/321

### DIFF
--- a/config/crd/bases/eda.ansible.com_edas.yaml
+++ b/config/crd/bases/eda.ansible.com_edas.yaml
@@ -1413,6 +1413,338 @@ spec:
                       type: object
                     type: array
                 type: object
+              scheduler:
+                default:
+                  replicas: 1
+                description: Defines desired state of eda-scheduler resources
+                properties:
+                  node_selector:
+                    additionalProperties:
+                      type: string
+                    description: NodeSelector for the EDA pods.
+                    type: object
+                  replicas:
+                    default: 2
+                    description: 'Size is the size of number of eda-scheduler replicas.
+                      Default: 1'
+                    format: int32
+                    minimum: 0
+                    nullable: true
+                    type: integer
+                  resource_requirements:
+                    description: Resource requirements for the eda-scheduler container.
+                    properties:
+                      claims:
+                        description: "Claims lists the names of resources, defined
+                          in spec.resourceClaims, that are used by this container.
+                          \n This is an alpha field and requires enabling the DynamicResourceAllocation
+                          feature gate. \n This field is immutable."
+                        items:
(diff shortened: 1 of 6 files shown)


rooftopcellist:
I deployed using aap-eda images built from this PR's branch (https://github.com/ansible/aap-eda/pull/263).  

Here is the spec I used:

```yaml
kind: EDA
metadata:
  name: eda-demo
spec:
  automation_server_url: https://awx-awx.apps.aap-dev.ocp4.testing.ansible.com
  automation_server_ssl_verify: 'no'
  service_type: ClusterIP
  ingress_type: Route
  no_log: false
  image_pull_policy: Always
  image_pull_secrets:
    - redhat-operators-pull-secret
  admin_password_secret: custom-admin-password
  image: quay.io/chadams/eda-server
  image_version: rq-scheduler
```

The deployment was successful and I was able to configure AWX and run a job via a rulebook activation in EDA.  For the activation pod and the automation job pod created and ran successfully.  

We should wait to merge this until the aap-eda/eda-server changes are in place.

config/crd/bases/eda.ansible.com_edas.yaml
```diff
@@ -1089,6 +1089,338 @@ spec:
...
+                  replicas: 1
+                description: Defines desired state of eda-scheduler resources
+                properties:
+                  node_selector:
+                    additionalProperties:
+                      type: string
+                    description: NodeSelector for the EDA pods.
+                    type: object
+                  replicas:
+                    default: 2
```
dsavineau, line 1103:
```suggestion
                    default: 1
```

rooftopcellist:
I will rebase and merged this change after https://github.com/ansible/eda-server-operator/pull/118 to make conflicts a little easier to deal with.

rooftopcellist:
After rebasing,  I re-deployed and set up a basic integration with my AWX instance.  The ActivationRulebook is successfully triggered, creating an activation-job:

image

You can also see that the scheduler pod is running as expected.  

The job ran successfully in AWX:

image

rooftopcellist:
Testing this with 5 rulebook activations shows that all are running as expected, triggering jobs in AWX.  

image

With 5 workers, when I push things a bit and create 10 RulebookActivations, we can see that only 5 will ever be in the running state at a given time, as expected with the current implementation.

image

rooftopcellist:
Looking at the AWX side, it is churning through the jobs that are created, but the pending jobs are hovering at around 5 with this load.  This could run away if we aren't careful, so improvements for throttling are needed.

PR #118, opened 2023-08-04
fix: correct worker into default worker and activation worker

kurokobo:
Closes #111 

Changes:

- Replace existing worker as default worker, and duplicate it as activation worker
- Modify args for `aap-eda-manage` in workers into `aap_eda.core.tasking.DefaultWorker` and `aap_eda.core.tasking.ActivationWorker`

I don't know what the service resource for worker is for since no service exists on [the manifests on eda-server repo](https://github.com/ansible/eda-server/tree/f7b2c73120dae4344288e74786d842d4049c6bc4/tools/deploy/eda-default-worker), but I create it for both workers to keep it as-is.

Note that this PR will cause conflicts with my other PRs, but since I don't know which PR will be approved, I'm keeping this in a way that it doesn't conflict with the `main` branch.
If conflicts arise in my PRs, including this, I will resolve them and push them, but please let me know if it is better to combine all (or specific) PRs into one PR.

Tested with:

```yaml
---
apiVersion: eda.ansible.com/v1alpha1
kind: EDA
metadata:
  name: eda
spec:
  ...
  api:
    replicas: 1
    resource_requirements:
      requests: {}
  ui:
    replicas: 1
    resource_requirements:
      requests: {}
  default_worker:
    replicas: 1
    resource_requirements:
      requests: {}
  activation_worker:
    replicas: 2
    resource_requirements:
      requests: {}
  redis:
    replicas: 1
    resource_requirements:
      requests: {}
```

Result:

```bash
$ kubectl -n eda logs deployments/eda-server-operator-controller-manager
...
----- Ansible Task Status Event StdOut (eda.ansible.com/v1alpha1, Kind=EDA, eda/eda) -----

PLAY RECAP *********************************************************************
localhost                  : ok=53   changed=0    unreachable=0    failed=0    skipped=17   rescued=0    ignored=0   
...

$ kubectl -n eda get pod,deployment,service
NAME                                                          READY   STATUS    RESTARTS   AGE
pod/eda-server-operator-controller-manager-6f974597f5-jgk6v   2/2     Running   0          5m9s
pod/eda-redis-7d78cdf7d5-npmvk                                1/1     Running   0          4m47s
pod/eda-postgres-13-0                                         1/1     Running   0          4m38s
pod/eda-ui-7cc5f59787-bpnsn                                   1/1     Running   0          3m49s
pod/eda-default-worker-59cd95f9d4-85wfc                       1/1     Running   0          3m44s
[...]

PR #119, opened 2023-08-11
Set default_worker replicas to 2 by default

rooftopcellist:
Follow-up to @kurokobo 's PR to add the default_worker group.  As discussed with @Alex-Izquierdo , this PR sets the default_worker's replicas to 2 by default.  

Related PR: 
* https://github.com/ansible/eda-server-operator/pull/118


rooftopcellist:
I now see 2 default-worker pods instead of 5.  

image

PR #122, opened 2023-08-15
Add logic for upgrading from Postgres 13 to 15

rooftopcellist:
##### SUMMARY

Upgrading from postgres 13 to 15.

Related PR:
* https://github.com/ansible/awx-operator/pull/1486

##### ISSUE TYPE

 - New or Enhanced Feature

### DIFF
--- a/roles/backup/vars/main.yml
+++ b/roles/backup/vars/main.yml
@@ -1,6 +1,6 @@
 ---
 deployment_type: "eda"
-_postgres_image: quay.io/sclorg/postgresql-15-c8s
+_postgres_image: quay.io/sclorg/postgresql-15-c9s
 _postgres_image_version: latest
 backup_complete: false
 database_type: "unmanaged"
--- a/roles/backup/vars/main.yml
+++ b/roles/backup/vars/main.yml
@@ -1,7 +1,7 @@
 ---
 deployment_type: "eda"
-_postgres_image: quay.io/sclorg/postgresql-13-c9s
+_postgres_image: quay.io/sclorg/postgresql-15-c8s
 _postgres_image_version: latest
 backup_complete: false
 database_type: "unmanaged"
-supported_pg_version: 13
+supported_pg_version: 15
--- a/roles/postgres/tasks/check_postgres_version.yml
+++ b/roles/postgres/tasks/check_postgres_version.yml
@@ -23,7 +23,7 @@
     set_fact:
       sorted_old_postgres_pods: "{{ filtered_old_postgres_pods |
         sort(attribute='metadata.name') |
-        reverse }}"
+        reverse | list }}"
     when: filtered_old_postgres_pods | length
 
   - name: Set info for previous postgres pod
--- a/roles/postgres/tasks/check_postgres_version.yml
+++ b/roles/postgres/tasks/check_postgres_version.yml
@@ -51,7 +51,7 @@
   block:
     - name: Set path to PG_VERSION file for given container image
       set_fact:
-        path_to_pg_version: '{{ combined_database.postgres_data_path }}/PG_VERSION'
+        path_to_pg_version: '{{ _postgres_data_path }}/PG_VERSION'
 
     - name: Get old PostgreSQL version
       k8s_exec:
--- a/roles/postgres/tasks/check_postgres_version.yml
+++ b/roles/postgres/tasks/check_postgres_version.yml
@@ -1,0 +1,76 @@
+---
+
+# It is possible that N-2 postgres pods may still be present in the namespace from previous upgrades.
+# So we have to take that into account and preferentially set the most recent one.
+- name: Get the old postgres pod (N-1)
+  k8s_info:
+    kind: Pod
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    field_selectors:
+      - status.phase=Running
+  register: _running_pods
+
+- block:
+  - name: Filter pods by name
+    set_fact:
+      filtered_old_postgres_pods: "{{ _running_pods.resources |
+        selectattr('metadata.name', 'match', ansible_operator_meta.name + '-postgres.*-0') |
+        rejectattr('metadata.name', 'search', '-' + supported_pg_version | string + '-0') |
+        list }}"
+
+  # Sort pods by name in reverse order (most recent PG version first) and set
+  - name: Set info for previous postgres pod
+    set_fact:
+      sorted_old_postgres_pods: "{{ filtered_old_postgres_pods |
+        sort(attribute='metadata.name') |
+        reverse }}"
+    when: filtered_old_postgres_pods | length
+
+  - name: Set info for previous postgres pod
+    set_fact:
+      old_postgres_pod: "{{ sorted_old_postgres_pods | first }}"
+    when: filtered_old_postgres_pods | length
+  when: _running_pods.resources | length
+
+- name: Look up details for this deployment
+  k8s_info:
+    api_version: "{{ api_version }}"
+    kind: "{{ kind }}"
+    name: "{{ ansible_operator_meta.name }}"
+    namespace: "{{ ansible_operator_meta.namespace }}"
+  register: this_eda
+
+# If this deployment has been upgraded before or if upgrade has already been started, set this var
+- name: Set previous PG version var
+  set_fact:
+    _previous_upgraded_pg_version: "{{ this_eda['resources'][0]['status']['upgradedPostgresVersion'] | default(false) }}"
+  when:
+    - "'upgradedPostgresVersion' in this_eda['resources'][0]['status']"
+
+- name: Check if postgres pod is running an older version
+  block:
+    - name: Set path to PG_VERSION file for given container image
+      set_fact:
+        path_to_pg_version: '{{ combined_database.postgres_data_path }}/PG_VERSION'
+
+    - name: Get old PostgreSQL version
+      k8s_exec:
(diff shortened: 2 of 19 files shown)


roles/backup/vars/main.yml
```diff
@@ -1,7 +1,7 @@
 ---
 deployment_type: "eda"
-_postgres_image: quay.io/sclorg/postgresql-13-c9s
+_postgres_image: quay.io/sclorg/postgresql-15-c8s
```
dsavineau, line 3: is there a reason to switch to the c8s image rather than sticking on the c9s one ?

roles/postgres/tasks/check_postgres_version.yml
```diff
@@ -0,0 +1,76 @@
+---
+
+# It is possible that N-2 postgres pods may still be present in the namespace from previous upgrades.
+# So we have to take that into account and preferentially set the most recent one.
+- name: Get the old postgres pod (N-1)
+  k8s_info:
+    kind: Pod
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    field_selectors:
+      - status.phase=Running
```
rcarrillocruz, line 10: Instead of getting old pods running in the namespace and then filter in next task, can't we just get the postgres pod by adding a field selector? There must be a label we can use?

roles/postgres/tasks/scale_down_deployment.yml
```diff
@@ -0,0 +1,23 @@
...
+
+- name: Check for presence of Deployment
+  k8s_info:
+    api_version: apps/v1
+    kind: Deployment
+    name: "{{ ansible_operator_meta.name }}-api"
+    namespace: "{{ ansible_operator_meta.namespace }}"
+  register: this_deployment
+
+# TODO: Determine if other deployments need to be scaled down too or not
```
dsavineau, line 11: I think activation workers now also have access to the database but it depends on the EDA version used

https://github.com/ansible/eda-server/blob/main/src/aap_eda/settings/default.py#L516-L518

roles/postgres/tasks/check_postgres_version.yml
rooftopcellist: We can't know for sure if the old postgres pod is version 12 or 13. That's why I did it this way.

roles/postgres/tasks/scale_down_deployment.yml
rooftopcellist: @dsavineau thanks!  I added in the 2 worker deployments and the scheduler deployments, all of which connect to the db. 

```
    - "{{ ansible_operator_meta.name }}-scheduler"
    - "{{ ansible_operator_meta.name }}-default-worker"
    - "{{ ansible_operator_meta.name }}-activation-worker"
 ```

**Comment by rooftopcellist** [comment]
I can't figure out why CI is failing when running [this line](https://github.com/ansible/galaxy-operator/blob/main/.ci/scripts/backup_and_restore.sh#L101):

```
pulp content list
```

@aknochow @dsavineau any ideas?

**Comment by dsavineau** [comment]
> I can't figure out why CI is failing when running [this line](https://github.com/ansible/galaxy-operator/blob/main/.ci/scripts/backup_and_restore.sh#L101):
> 
> ```
> pulp content list
> ```
> 
> @aknochow @dsavineau any ideas?

I think this comment should have landed on the galaxy-operator repository instead of the eda operator right ?

**Comment by dsavineau** [comment]
not sure if I'm testing this correctly because this should be the same code than awx-operator but the upgrade fails when searching for the previous postgresql pod.

```console
TASK [Set info for previous postgres pod] **************************************
task path: /opt/ansible/roles/postgres/tasks/check_postgres_version.yml:22

ok: [localhost] => {
    "ansible_facts": {
        "sorted_old_postgres_pods": "<list_reverseiterator object at 0x7f539eaa5610>"
    },
    "changed": false
}

TASK [Set info for previous postgres pod] **************************************
task path: /opt/ansible/roles/postgres/tasks/check_postgres_version.yml:29

        "old_postgres_pod": "<"
    },
    "changed": false
}
```
Also I noticed that `postgres_keep_pvc_after_upgrade` is used but not defined in defaults nor the CRD

dsavineau:
alright this seems related to the ansible version (and/or python version) used by the operator

- EDA operator uses v1.27.0
  - ansible 2.9.27
  - python 3.8

- AWX operator uses v1.34.0
  - ansible-core 2.15.8
  - python 3.9

if I switch to v1.34.0 then I don't have the issue anymore. I'll need to look more to understand why it's failing with the old version because we're likely to still use ansible 2.9 downstream

dsavineau:
The reverse filter in ansible 2.9 returns an iterator and we cast it to list to avoid this issue.
With ansible-core 2.15 we don't have to do that but keeping the `list` filter isn't an issue.
I would say that we should add the `list` filter even if we plan to update the operator sdk to a newer image because ansible 2.9 is still use for downstream

dsavineau:
@rooftopcellist I've pushed some fixes for this PR so don't forget to fetch the changes if you want to push new things

With those fixes, the error I'm facing occurs during the postgresql dump/restore on the pg 15 pod

```console
pg_restore: error: could not execute query: ERROR:  must be member of role "postgres"
Command was: ALTER SCHEMA public OWNER TO postgres;

pg_restore: warning: errors ignored on restore: 1
```

Note sure why we have this. I'll continue to debug it next week and try to do a similar pg upgrade for the awx operator to see if we don't have the same symptoms

dsavineau:
@rooftopcellist with the latest changes I pushed, I was able to upgrade from 13 to 15 without issue.

However, the change about granting the postgres role temporary to the eda postgresql user seems weird to me because we never had to do this in other operators for other postgresql upgrade (12 or 13)

https://github.com/ansible/eda-server-operator/pull/122/commits/c2b13fd4d1d1facc8f0d8571a2111e7e085e96a7

kurokobo:
https://github.com/ansible/galaxy-operator/pull/80#issuecomment-2022065152

dsavineau:
> https://github.com/ansible/galaxy-operator/pull/80#issuecomment-2022065152

I have less concerns about the postgresql upgrade from 13 to 15 for the EDA operator because it was already using postgresql image from sclorg as default.

rooftopcellist:
@kurokobo I tested out a fresh install on k3s following similar steps to what you've shown in your awx-on-k3s repo and was not able to reproduce the permission error in the postgres pod that caused us to add these changes to the awx-operator:
* https://github.com/ansible/awx-operator/pull/1805

I pinned the redis image to `c9s` and will cut an eda-server-operator 1.0.2 release so that users have a working release before the PostgreSQL 15 upgrade (which we will likely put out as a 2.0.0 release). @kurokobo I plan to wait on releasing the PG15 changes until you and/or some other community folks get a chance to try out these.  

Since all of the comments have been addressed and tests are passing in every scenario I can think of, I plan to merge this shortly. 

## Testing

**On Openshift:**

I ran through a fresh install, backup, restore, and upgraded from `main` --> the `pg15` branch all without error.  

**On k3s:**
I did a fresh install, backup, restore and upgrade with no operator errors and no errors in the logs of the resulting deployments.  

Notes for follow-up:
I noticed two things about how the pods scale down before upgrading the database:
* Currently, it loops through deployments and scales them down one by one, which is slow.  It would be faster to scale down all of the deployments at once if possible.  But this is only for postgresql upgrades, which are rare, so I'm inclined to leave it the way it is.
* we may want to scale down ui and redis as well while upgrade (just so that errors don't show in the eda-ui pod.

kurokobo:
@rooftopcellist 
Thanks, I will test 1.0.2 first, then test `main` branch, and let you know the results.
F.Y.I., I don't know if you know this but my repository already contains the guide to deploy EDA Server using EDA Server Operator. I've added this the last summer when I started contribution for EDA Server Operator : https://github.com/kurokobo/awx-on-k3s/tree/main/rulebooks

rooftopcellist:
I hadn't seen that yet, thanks for the link!  I'll check out the similar one for galaxy.  Thanks for checking out the changes!

kurokobo:
@rooftopcellist 
Sorry for my delay, and sorry for adding a comment on closed PR.
I've tested for both 1.0.2 and `main` branch and I can confirm that both work as expected in senarios for fresh installation and upgrading.

One concern is that the `postgres_data_volume_init` implemented in AWX Operator is not implemented in EDA Server Operator.
As long as we use sclorg's PSQL, the permission error for UID:26 that occurred in AWX Operator can also occur in this Operator if users are using specific backend for PVs, e.g. hostPath, Longhorn, etc..
I suspect that the reason you could not reproduce it in your environment was because you reused the directory you created for AWX with `chown`ed to `26:0`.

To reproduce, please follow my guide again, with skipping `sudo chown 26:0 /data/eda/postgres-13/data`. Currently my guide deployes 1.0.2: https://github.com/kurokobo/awx-on-k3s/blob/main/rulebooks/README.md

I think it is a good idea to implement `postgres_data_volume_init` at this time, as the same issues will probably be reported as the number of users grows. I guess the reason why this is not reported yet is simply there are not many users for this Operator. Also, if the users uses my guide to deploy EDA Server, they would not faced this issue since I already added `chown` for UID:26 to the guide from the beginning.

Any thought?

rooftopcellist:
@kurokobo ahh you are absolutely correct.  I did have the chown 26 line in my reproducer script because I saw it in your repo instructions.  

But I agree, it's better to implement `postgres_data_volume_init` here as well, then users won't need that extra line for k3s hostPath and longhorn, etc.

PR #133, opened 2023-09-20
Add back the EDA_SERVER env var

rooftopcellist:
Backstory here:
* https://github.com/ansible/eda-server/pull/412

But we are adding back the `EDA_SERVER` env var and in the future will remove `EDA_PROTOCOL` and `EDA_HOST` in the future.  Count them as deprecated for now.


rooftopcellist:
Hold off on merging this until https://github.com/ansible/eda-server/pull/412 lands.

jamestalton (approved):
lgtm

PR #143, opened 2023-11-02
Add X-Forwarded headers to nginx.conf

rooftopcellist:
These headers should also be included.


roles/eda/templates/eda.configmap.yaml.j2
```diff
@@ -72,11 +72,22 @@ data:
             location ~ ^/api/eda/v[0-9]+/ {
                 proxy_pass http://{{ ansible_operator_meta.name }}-api:8000;
                 proxy_set_header Origin http://{{ ansible_operator_meta.name }}-api:8000;
+                proxy_set_header Host $http_host;
```
Alex-Izquierdo, line 75: I did the copy paste from our nginx in eda-server, but I'm not sure if are all the right values. Let me confirm with what we have in downstream.

Alex-Izquierdo: I think that's fine.

Alex-Izquierdo (approved):
LGTM

PR #151, opened 2023-11-17
Respect the deprecated worker parameters as a fallack 

rooftopcellist:
Respect the deprecated worker parameters as a fallack if new worker params are not set. This adds backwards compatibility for users upgrading from versions of the eda-server-operator that only had the single "worker" deployment, before default-worker and activation-worker deployments were added.

### DIFF
--- a/roles/eda/tasks/combine_defaults.yml
+++ b/roles/eda/tasks/combine_defaults.yml
@@ -1,9 +1,33 @@
 ---
 
-- name: Combine default and custom variables from the CR for each component
+- name: Combine default variables for components
   set_fact:
-    combined_api: "{{ _api | combine (api, recursive=True ) }}"
-    combined_ui: "{{ _ui | combine (ui, recursive=True) }}"
-    combined_default_worker: "{{ _default_worker | combine (default_worker, recursive=True) }}"
-    combined_activation_worker: "{{ _activation_worker | combine (activation_worker, recursive=True) }}"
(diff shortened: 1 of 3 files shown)


rcarrillocruz:
Are we looping docs team on this one?

roles/eda/tasks/combine_defaults.yml
```diff
@@ -1,9 +1,34 @@
...
+- name: Set defaults for workers # (overridden by worker, default_worker, and activation_worker)
+  set_fact:
+    combined_default_worker: "{{ _default_worker }}"
+    combined_activation_worker: "{{ _activation_worker }}"
+
+- name: Combine worker params
+  set_fact:
+    combined_worker: "{{ _worker | combine (worker, recursive=True) }}"
+    combined_default_worker: "{{ combined_worker}}"
+    combined_activation_worker: "{{ combined_worker}}"
```
dsavineau, line 19: this won't work because you can't reuse a fact defined in the same set_fact task so combined_worker either 
- needs to be set in a separate task before the one for combined_default_worker and combined_default_activation
- needs to duplicate the combined_worker value for the two other facts
```suggestion
    combined_default_worker: "{{ _worker | combine (worker, recursive=True) }}"
    combined_activation_worker: "{{ _worker | combine (worker, recursive=True) }}"
```

rooftopcellist: Thanks!  I updated the PR so that it just does it in line so that we have one fewer task.

dsavineau (approved):

PR #152, opened 2023-12-12
Redirect traffic to eda/ urlpath in nginx config

rooftopcellist:
Closes: https://github.com/ansible/eda-server-operator/issues/153

The eda-ui container now serves up the static files at the /eda/ path, the operator's nginx.conf configmap needs to be updated to reflect this, otherwise the UI shows as blank.  

I also added a bit to redirect `/eda` --> `/eda/`, without it, users may see a 404.  

cc @kurokobo as you have probably seen this issue.

### DIFF
--- a/roles/eda/templates/eda.configmap.yaml.j2
+++ b/roles/eda/templates/eda.configmap.yaml.j2
@@ -87,10 +87,15 @@ data:
             }
 
             location / {
+                # Redirect all traffic from root to /eda/
+                return 301 $scheme://$host/eda/;
+            }
+
+            location /eda/ {
                 autoindex off;
                 expires off;
                 add_header Cache-Control "public, max-age=0, s-maxage=0, must-revalidate" always;
-                try_files $uri /index.html =404;
+                try_files $uri $uri/ /index.html =404;
             }
-          }
-    }
+        }
+      }
--- a/roles/eda/templates/eda-ui.deployment.yaml.j2
+++ b/roles/eda/templates/eda-ui.deployment.yaml.j2
@@ -64,13 +64,13 @@ spec:
         - containerPort: 8080
         readinessProbe:
           httpGet:
-            path: /
+            path: /eda/
             port: 8080
           failureThreshold: 1
           periodSeconds: 10
         livenessProbe:
           httpGet:
-            path: /
+            path: /eda/
             port: 8080
           failureThreshold: 1
           periodSeconds: 10


roles/eda/templates/eda.configmap.yaml.j2
```diff
@@ -87,10 +87,20 @@ data:
...
+                return 301 $scheme://$host/eda/;
+            }
+
+            location /eda/ {
                 autoindex off;
                 expires off;
                 add_header Cache-Control "public, max-age=0, s-maxage=0, must-revalidate" always;
-                try_files $uri /index.html =404;
+
+                # Serve /usr/share/nginx/html/index.html for requests to /eda/
```
Alex-Izquierdo, line 102: is this comment correct?

roles/eda/templates/eda.configmap.yaml.j2
```diff
@@ -87,10 +87,20 @@ data:
             }
 
             location / {
+                rewrite ^/$ /eda/ permanent; # Redirects all traffic from root to /eda/
```
Alex-Izquierdo, line 90: Why not use here `return 301 $scheme://$host/eda/;` as well

roles/eda/templates/eda.configmap.yaml.j2
jamesmarshall24: According to the config above, it looks like it:
```
root /usr/share/nginx/html;
```

rooftopcellist: It is, however, downstream the root path changes, so probably better to just leave it out anyways.

roles/eda/templates/eda.configmap.yaml.j2
rooftopcellist: Good point, I've updated the PR.  Thanks!

Alex-Izquierdo: ok, but one of both, or you do the rewrite or the redirect. `return 301 $scheme://$host/eda/;` I would say is simpler and efficient.

kurokobo:
> cc @kurokobo as you have probably seen this issue.

Nope I didn't have fresh deployment in this few weeks, thanks for identifying and working on this!

Alex-Izquierdo (approved):

roles/eda/templates/eda.configmap.yaml.j2
rooftopcellist: Thank you for clarifying, I just updated it.  In testing, I am able to navigate to the UI and it redirects correctly.

rooftopcellist:
I just created a new 0.0.11 tag and pushed images for it, so that users have a commit pointer to install from.

Issue #154, opened 2023-12-13
On fresh deployments, sometimes the first reconciliation loop fails on the task "Check for API Pod"


kurokobo:
#### Description

On fresh deployments for EDA CR, sometimes the first reconciliation loop fails on the task "Check for API Pod".

If we do nothing and wait for the next reconciliation loop, this task will succeed and the deployment will complete.

The first reconciliation:

```bash
...
--------------------------- Ansible Task StdOut -------------------------------
 TASK [Check for API Pod] ******************************** 
fatal: [localhost]: FAILED! => {"api_found": true, "attempts": 3, "changed": false, "resources": []}
...
[eda : Check for API Pod] *************************************************\r\n
task path: /opt/ansible/roles/eda/tasks/deploy_eda.yml:31\n
FAILED - RETRYING: Check for API Pod (3 retries left).\n
FAILED - RETRYING: Check for API Pod (2 retries left).\n
FAILED - RETRYING: Check for API Pod (1 retries left).\n
fatal: [localhost]: FAILED! => {\"api_found\": true, \"attempts\": 3, \"changed\": false, \"resources\": []}\n\r\n
PLAY RECAP *********************************************************************\r\n
localhost                  : ok=53   changed=7    unreachable=0    failed=1    skipped=12   rescued=0    ignored=0   \r\n\n
```

Just wait for the second reconciliation with doing nothing:

```bash
...
TASK [eda : Check for API Pod] *************************************************
task path: /opt/ansible/roles/eda/tasks/deploy_eda.yml:31

-------------------------------------------------------------------------------
{"level":"info","ts":1702472703.452893,"logger":"proxy","msg":"cache miss: /v1, Kind=PodList err-Index with name field:status.phase does not exist"}
{"level":"info","ts":1702472703.856233,"logger":"logging_event_handler","msg":"[playbook task start]","name":"eda","namespace":"eda","gvk":"eda.ansible.com/v1alpha1, Kind=EDA","event_type":"playbook_on_task_start","job":"4939066471100038729","EventData.Name":"eda : Create admin user"}

--------------------------- Ansible Task StdOut -------------------------------

TASK [eda : Create admin user] *************************************************
task path: /opt/ansible/roles/eda/tasks/main.yml:36
```

#### Additional information

This is due to the `until` without `delay` or `retries`.
It wait only for 15 seconds since the default value for `delay` is `5` and `retries` is `3`. 

[...]

PR #157, opened 2023-12-13
Increase how long we wait for the api pod to come up

rooftopcellist:
Resolves: https://github.com/ansible/eda-server-operator/issues/154

This PR increases how long we wait for the api pod to come up
- There are 3 initContainers and migrations that need to run before the API pod is in the running state, this is expected to take a bit of time.
- This will avoid the reconciliation loop failing when checking the API pod (default was only 15 seconds prior to this change).

cc @kurokobo

### DIFF
--- a/roles/eda/tasks/deploy_eda.yml
+++ b/roles/eda/tasks/deploy_eda.yml
@@ -44,6 +44,9 @@
     - "eda_api_pod['resources'] | length"
     - "eda_api_pod['resources'][0]['status']['phase'] == 'Running'"
     - "eda_api_pod['resources'][0]['status']['containerStatuses'][0]['ready'] == true"
+    - "eda_api_pod['resources'][0]['status']['containerStatuses'][1]['ready'] == true"
+  retries: 60
+  delay: 5
 
 - name: Set the resource pod name as a variable.
   set_fact:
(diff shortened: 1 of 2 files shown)


dsavineau (approved):

kurokobo:
Thanks for working on this!
I have confirmed that it works as expected; the first loop completes without failure.

Just to confirm, is it the expected behavior that the task succeeds and moves on even if the pod is `1/2 Running`?

- `eda_api_pod['resources'][0]['status']['containerStatuses'][0]` appears to be `daphne`, not `eda-api` (`containerStatuses` are ordered by name).
- In the current implementation, if `daphne` instead of `eda-api` becomes `ready: true`, the task succeeds.
- The startup of `daphne` completes immediately, but `eda-api` will not be `ready: true` for at least 90 seconds because of the `readinessProbe` with `initialDelaySeconds: 90`.

Should we wait until it is `2/2 Running` before proceeding like this? Also `initialDelaySeconds` could be a little shorter.

```yaml
  until:
    - "eda_api_pod['resources'] | length"
    - "eda_api_pod['resources'][0]['status']['phase'] == 'Running'"
    - "eda_api_pod['resources'][0]['status']['containerStatuses'][0]['ready'] == true"
    - "eda_api_pod['resources'][0]['status']['containerStatuses'][1]['ready'] == true"     👈👈👈
```

Again, even if we proceed at `1/2 Running` (only `daphne` is ready), the deployment will complete without any issues, since probably the startup of `eda-api` is completed internally within 90 seconds.

rooftopcellist:
Thank you for the feedback and suggestions @kurokobo!

I reduced the initialDelaySeconds to 20 seconds and added the bit to wait for both containers of the api pod.  Thanks!

Issue #125, opened 2023-08-17
Pin API and UI image refs in released operator images


rooftopcellist:
There are no gaurantees that newer api and ui images will be compatible with a given released eda-server-operator image.  We should pin the latest api and ui image at time of release inside the operator image as env vars.  

Here is how it was done in AWX Operator:
* https://github.com/ansible/awx-operator/pull/612/files

We will need to coordinate with the eda-server and eda-ui image maintainers to standardize tags as needed.  We will also want to write a release script or GHA that accepts a version or image tags as inputs.

Daniel-Vaz:
Regarding this issue I think that also currently a high priority is make sure the `eda-server-operator` has a tag that is stable and is working with the latest `eda-ui` and `eda-server` images. 

From my testing currently in order to not use the "latest"\"main" versions of all these 3 images, we can only use the `eda-server-operator` tag `0.0.4`, and because of these issues ([UI ](https://github.com/ansible/eda-server-operator/issues/123#issue-1852457286)\ [Server](https://github.com/ansible/eda-server-operator/issues/111#issuecomment-1665232248)), we can only use the image older and specific image tags.

Based on the recent worker CRD changes I believe it would be crucial to have a version `0.0.5` of the eda-server-operator stable enough with some statically defined `eda-server` and `eda-ui` working\compatible images.

goldyfruit:
Just got a similar issue with tag `0.0.10`, by default the `latest` tag is tried but doesn't exist.

Here is a quick workaround to get EDA deployed.

```yaml
  image: quay.io/ansible/eda-server
  image_version: main
  image_web: quay.io/ansible/eda-ui
  image_web_version: main
```

iwt-cmd:
> Just got a similar issue with tag `0.0.10`, by default the `latest` tag is tried but doesn't exist.
> 
> Here is a quick workaround to get EDA deployed.
> 
> ```yaml
>   image: quay.io/ansible/eda-server
>   image_version: main
>   image_web: quay.io/ansible/eda-ui
>   image_web_version: main
> ```

We ran into the same issue with 0.0.11.  The "eda-ui" repo has been updated with a "latest" tag but the "eda-server" is still missing it.  We used the "image" and "image_version" flags mentioned above as a workaround.

kurokobo:
For both `eda-server:main` and `eda-ui:latest`, backward-incompatible changes may often occur and it causes a previously successful deployment to fail in the future.
[...]

Issue #103, opened 2023-07-10
Enable single command install without clone


rooftopcellist:
There's currently no single command apply for any of the 3 operators that doesn't require cloning the repo.  This is how other operators do it, which makes it easier for people to get up and running faster. Example: Installing the standard kubernetes metrics:

```
kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/latest/download/components.yaml
```

We can add it to a release GHA, then have it be a .yaml release artifact.  That's how they do it for the metrics-server:

* make command to generate the manifests
* step in Release GHA to publish the artifacts - https://github.com/kubernetes-sigs/metrics-server/blob/796fc0f832c1ac444c44f88a952be87524456e07/.github/workflows/release.yaml#L24

PR #159, opened 2024-01-04
Add Release Workflow and Single Command Install artifact

rooftopcellist:
This PR adds a make target to generate a single yaml file that can be applied to install the operator for a particular version.

This yaml file will be added as a release artifact for operator releases.

This PR also adds a GHA for automating releases.

Related issues this will resolve:
* https://github.com/ansible/eda-server-operator/issues/103
* https://github.com/ansible/eda-server-operator/issues/125


rooftopcellist:
I tested this out on my fork and it now works.  Here is an example of a draft release:
* https://github.com/rooftopcellist/eda-server-operator/releases

Things that will be needed:
* the current QUAY_TOKEN will need to grow permissions for the bundle and catalog quay.io registries.

rooftopcellist:
We should double check that the one line operator.yaml artifact deploys a working eda-server-operator when applied. Then we should be sure to document this install method to make users aware of it.  We should include how to upgrade using the one line install too and test that.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,102 @@
...
+          docker tag eda-server-operator:${{ github.event.inputs.version }} eda-server-operator:latest
+
+      - name: Push Operator Image
+        uses: redhat-actions/push-to-registry@v2.1.1
+        with:
+          image: eda-server-operator
+          tags: ${{ github.event.inputs.version }}
+          registry: quay.io/ansible/
+          username: ${{ secrets.QUAY_USER }}
+          password: ${{ secrets.QUAY_TOKEN }}
```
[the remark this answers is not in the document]
rooftopcellist: I think this will work nicely, we will just have to make sure that the GHA that builds and pushes the `main` branch image has completed before running the release GHA, which shouldn't really be a problem.  

@jon-nfc I will make the changes you suggested in the [other PR](https://github.com/ansible/eda-server-operator/pull/161) for multi-arch builds as well.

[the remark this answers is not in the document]
jon-nfc: Good pickup I didn't even consider a possible collision of not being ready. Although TBH the builds are pretty fast.

further context on the reasoning of why i chose this method is under heading 'Additional info' in ansible/awx-operator/pull/1681

rooftopcellist:
I tried running this on my fork and it works as expected, but fails on the bundle push atm.  That is because I need to get the permissions added tot he quay token, once done, we can merge this and put out our first stable major version of the operator, v1.0.0!

rooftopcellist:
@jon-nfc Unfortunately, since we need to be able to pin the version of the `eda-server` and `eda-ui` images at build time of the operator, that poses a problem for this approach of using the pre-built images from devel, re-tagging, and releasing them...

With that in mind, unless you have any other ideas, I think we should revert the last commit, as the next change I intend to make is to add pinned default versions via env vars to resolve this issue:
* https://github.com/ansible/eda-server-operator/issues/125

.github/workflows/release.yaml
```diff
@@ -0,0 +1,118 @@
...
+
+      - name: Install operator-sdk
+        run: |
+          OPERATOR_SDK_VERSION=v1.23.0
+          curl -Lo operator-sdk "https://github.com/operator-framework/operator-sdk/releases/download/${OPERATOR_SDK_VERSION}/operator-sdk_linux_amd64"
+          chmod +x operator-sdk
+          sudo mv operator-sdk /usr/local/bin/
```
dsavineau, line 66: do we need this since the bundle target already call the operator-sdk target which does the same job ?
Also the operator sdk version in this task and the one defined in the Makefile is different. Is there a reason for that ?

jon-nfc:
> @jon-nfc Unfortunately, since we need to be able to pin the version of the `eda-server` and `eda-ui` images at build time of the operator, that poses a problem for this approach of using the pre-built images from devel, re-tagging, and releasing them...
> 
> With that in mind, unless you have any other ideas, I think we should revert the last commit, as the next change I intend to make is to add pinned default versions via env vars to resolve this issue:
> 
> * [Pin API and UI image refs in released operator images #125](https://github.com/ansible/eda-server-operator/issues/125)
> 
> I went ahead and reverted it for now, but it is still open for discussion.

is the work from  https://github.com/ansible/awx-operator/pull/1681#discussion_r1445694531 portable to this repo? Although chain of build does change, it would remain the same for the release cycle as the cycle starts at "staging"

Makefile
```diff
@@ -137,6 +137,22 @@ KUSTOMIZE = $(shell which kustomize)
...
 
+.PHONY: operator-sdk
+OPERATOR_SDK = $(shell pwd)/bin/operator-sdk
+operator-sdk: ## Download operator-sdk locally if necessary, preferring the $(pwd)/bin path over global if both exist.
+ifeq (,$(wildcard $(OPERATOR_SDK)))
+ifeq (,$(shell which operator-sdk 2>/dev/null))
+	@{ \
+	set -e ;\
+	mkdir -p $(dir $(OPERATOR_SDK)) ;\
+	curl -sSLo $(OPERATOR_SDK) https://github.com/operator-framework/operator-sdk/releases/download/v1.32.0/operator-sdk_$(OS)_$(ARCHA) ;\
```
dsavineau, line 148:
```suggestion
	curl -sSLo $(OPERATOR_SDK) https://github.com/operator-framework/operator-sdk/releases/download/v1.32.0/operator-sdk_$(OS)_$(ARCH) ;\
```

rooftopcellist: we don't have helm releases and thus don't need kube-slice in the eda-server-operator.  So best to just change it to ARCH like you mentioned.

.github/workflows/release.yaml
rooftopcellist: I have removed this task now that the typo is fixed, thanks!

.github/workflows/release.yaml
```diff
@@ -0,0 +1,85 @@
...
+      - name: Pull Stored Image and Publish to quay.io/${{ github.repository }}:latest
+        run: |
+          docker buildx imagetools create \
+            ghcr.io/${{ github.repository_owner }}/${{ github.repository }}:${TAG_NAME} \
+            --tag quay.io/ansible/${{ github.repository }}:latest
+
+      - name: Build Bundle Image
+        run: |
+          make bundle IMG=eda-server-operator:${{ github.event.inputs.version }} VERSION=${{ github.event.inputs.version }}
+          make bundle-build BUNDLE_IMG=eda-server-operator-bundle:${{ github.event.inputs.version }}
```
jon-nfc, line 61: You'll have to forgive my unfamiliarity....

I'm assuming that the docker file in this step is created as part of this line, in the bundle target?

https://github.com/ansible/eda-server-operator/blob/f237b95b030af8308c8373f6cf11dbc19f094566/Makefile#L176

README.md
```diff
@@ -195,3 +213,10 @@ spec:
 - [No Log](./docs/user-guide/advanced-configuration/no-log.md)
 - [Deploy a Specific Version of EDA](./docs/user-guide/advanced-configuration/deploying-a-specific-version.md)
 - [Trusting a Custom Certificate Authority](./docs/user-guide/advanced-configuration/trusting-a-custom-certificate-authority.md)
+
+## Maintainers Docs
+
+Maintainers of this repo need to carry out releases, triage issues, etc. There are docs for those types of administrative tasks in the `docs/maintainer/` directory.
```
jon-nfc, line 219: Food for thought, maintainer and contributing docs be moved to a central location.

Thought Process. I know of at least three repos in the ansible org where the structures and processes appear the same:

- awx
- awx-operator
- eda-operator

A single location for common items give a single point for them to be updated and referenced by all. Normalizing the processes, i.e Actions, Workflows, Contributing, Dev etc. Reduces the burden on those who work on the repos as now they need to learn one process to work on "Ansible" repos and the process they learn is portable to all repos. An update to the source docs, should add a workflow item to review and refactor a repos "processes" to be in line and ensures it remains up to date and inline with best practices.

docs/user-guide/advanced-configuration/deploying-a-specific-version.md
```diff
@@ -5,9 +5,9 @@ There are a few variables that are customizable for eda the image management.
 | Name                   | Description               | Default                                 |
 | ---------------------- | ------------------------- | --------------------------------------  |
 | image                  | Path of the image to pull | quay.io/ansible/eda-server              |
-| image_version          | Image version to pull     | latest                                  |
+| image_version          | Image version to pull     | value of DEFAULT_EDA_VERSION or main    |
 | image_web              | Path of the image to pull | quay.io/ansible/eda-ui                  |
-| image_web_version      | Image version to pull     | latest                                  |
+| image_web_version      | Image version to pull     | value of DEFAULT_EDA_UI_VERSION or main |
```
jon-nfc, line 10: the tags in the builds as part of release use the `latest` tag

roles/eda/defaults/main.yml
```diff
@@ -12,9 +12,10 @@ image_pull_secrets: []
 
 _image: quay.io/ansible/eda-server
 _image_version: main
+_image_version: "{{ lookup('env', 'DEFAULT_EDA_VERSION') or 'main' }}"
 
 _image_web: quay.io/ansible/eda-ui
-_image_web_version: main
+_image_web_version: "{{ lookup('env', 'DEFAULT_EDA_UI_VERSION') or 'main' }}"
```
jon-nfc, line 18: the tags in the builds as part of release use the `latest` tag

jon-nfc (commented):
done. only issue I noticed was inconsistency between `main` and `latest` image tags

README.md
rooftopcellist: @jon-nfc I :100: agree.  We an internal repo where we could potentially put standard release docs like that, but no public equivalent.  I really want to keep these docs public.  

A step in the right direction would be to have a single set of docs for Actions, Workflows, Contributing topics in one of those mentioned repos (maybe awx?), then to link to that from the other repo's readme's.  However, having a central place would be better, as a landing page for people interested in contributed to any of the Ansible related repos.  

The developer docs will still be largely different for each repo, as developing for say ansible-core, will be vastly different then awx, or any of the operators.  But I could certainly see a single development.md doc for all of the operators.  

There is leg work needed to standardize these processes in each operator repo, and arguably more to be done in the application repos (awx, eda, galaxy_ng), but it is definitely worth the time.  I think we should capture this in a separate issue.

docs/user-guide/advanced-configuration/deploying-a-specific-version.md
rooftopcellist: @jon-nfc the `image_version` and `image_web_version` parameters are referring to the tags for the application images ([eda-server](quay.io/ansible/eda-server) and [eda-ui](quay.io/ansible/eda-ui)), not the operator.  

The app images params default to `main` in the development branch (`main`), but in released operators, they will be pinned to the version of the application image tested at the time the operator was released.

So we need to stick with `main` for the app images. Also, the eda-server repo does not have a `latest` tag.

roles/eda/defaults/main.yml
rooftopcellist: See above [comment](https://github.com/ansible/eda-server-operator/pull/159#discussion_r1447492311).

rooftopcellist:
With the latest set of changes, the release workflow (stage and promote) are now both working!

* [successful stage workflow run](https://github.com/rooftopcellist/eda-server-operator/actions/runs/7481455238)
* [successful promote workflow run](https://github.com/rooftopcellist/eda-server-operator/actions/runs/7482060697)
* [New tag for test release 0.0.114](https://github.com/rooftopcellist/eda-server-operator/releases/tag/0.0.114)

With that, this is ready to merge.
cc @dsavineau

.github/workflows/release.yaml
```diff
@@ -0,0 +1,112 @@
...
+        with:
+          registry: ${{ github.event.inputs.quay_registry }}
+          username: ${{ secrets.QUAY_USER }}
+          password: ${{ secrets.QUAY_TOKEN }}
+
+      - name: Pull Stored Image and Publish to ${{ github.event.inputs.quay_registry }}/eda-server-operator:${TAG_NAME}
+        run: |
+          docker buildx imagetools create \
+            ghcr.io/${{ github.repository }}:${TAG_NAME} \
+            --tag ${{ github.event.inputs.quay_registry }}/eda-server-operator:${TAG_NAME}
```
dsavineau, line 55: Not a blocker for this PR but I'm pretty sure that docker buildx build supports using multiple tag at the same time but not sure about docker buildx imagetools create (if one wants to try it)
This would allow to combine both tasks with 

```suggestion
          docker buildx imagetools create \
            ghcr.io/${{ github.repository }}:${TAG_NAME} \
            --tag ${{ github.event.inputs.quay_registry }}/eda-server-operator:${TAG_NAME}
            --tag ${{ github.event.inputs.quay_registry }}/eda-server-operator:latest
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,112 @@
...
+        run: |
+          docker buildx imagetools create \
+            ghcr.io/${{ github.repository }}:${TAG_NAME} \
+            --tag ${{ github.event.inputs.quay_registry }}/eda-server-operator:latest
+        working-directory: eda-server-operator
+
+      - name: Build Bundle Image
+        run: |
+          make bundle IMG=eda-server-operator:${{ github.event.inputs.tag_name }} VERSION=${{ github.event.inputs.tag_name }}
+          make bundle-build BUNDLE_IMG=eda-server-operator-bundle:${{ github.event.inputs.tag_name }}
```
dsavineau, line 68: is there any reason not using both bundle and bundle-build targets with a single make command ?
```suggestion
          make bundle bundle-build IMG=eda-server-operator:${{ github.event.inputs.tag_name }} VERSION=${{ github.event.inputs.tag_name }} BUNDLE_IMG=eda-server-operator-bundle:${{ github.event.inputs.tag_name }}
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,112 @@
...
+          echo "TAG_NAME=${{ github.event.release.tag_name }}" >> $GITHUB_ENV
+
+      - uses: actions/checkout@v3
+        with:
+          ref: 'main'
+          depth: 0
+          path: eda-server-operator
+
+      - name: Log into registry ghcr.io
+        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d    # v3.0.0
```
dsavineau, line 37: Just curious about why we're using the sha1 rather than the version (either short or full one)
```suggestion
        uses: docker/login-action@v3
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,112 @@
...
+
+      - name: Build Bundle Image
+        run: |
+          make bundle IMG=eda-server-operator:${{ github.event.inputs.tag_name }} VERSION=${{ github.event.inputs.tag_name }}
+          make bundle-build BUNDLE_IMG=eda-server-operator-bundle:${{ github.event.inputs.tag_name }}
+          docker tag eda-server-operator-bundle:${{ github.event.inputs.tag_name }} eda-server-operator-bundle:latest
+        working-directory: eda-server-operator
+
+      - name: Push Bundle Image
+        uses: redhat-actions/push-to-registry@v2.1.1
```
dsavineau, line 73: Any reason not use the latest version here ? 2.1.1 is 3 years old
```suggestion
        uses: redhat-actions/push-to-registry@v2.7.1
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,112 @@
...
+          password: ${{ secrets.QUAY_TOKEN }}
+
+      - name: Push Bundle Image
+        uses: redhat-actions/push-to-registry@v2.1.1
+        with:
+          image: eda-server-operator-bundle
+          tags: latest
+          registry: quay.io/ansible/
+          username: ${{ secrets.QUAY_USER }}
+          password: ${{ secrets.QUAY_TOKEN }}
```
dsavineau, line 88: The push-to-registry action supports using multiple tags so no need to have two tasks for this

```suggestion
          image: eda-server-operator-bundle
          tags: ${{ github.event.inputs.tag_name }} latest
          registry: quay.io/ansible/
          username: ${{ secrets.QUAY_USER }}
          password: ${{ secrets.QUAY_TOKEN }}
```

https://github.com/redhat-actions/push-to-registry#image-tag-inputs

.github/workflows/release.yaml
```diff
@@ -0,0 +1,112 @@
...
+
+      - name: Push Catalog Image
+        uses: redhat-actions/push-to-registry@v2.1.1
+        with:
+          image: eda-server-operator-catalog
+          tags: latest
+          registry: quay.io/ansible/
+          username: ${{ secrets.QUAY_USER }}
+          password: ${{ secrets.QUAY_TOKEN }}
```
dsavineau, line 112: ditto

dsavineau (approved):
some comments and minor improvements but could be done as a follow up PR

.github/workflows/release.yaml
jon-nfc: explained here https://github.com/ansible/eda-server-operator/pull/161#discussion_r1445670410

.github/workflows/release.yaml
rooftopcellist: I had them comma-separated before, which didn't work.  Thanks for  pointing this out.

.github/workflows/release.yaml
rooftopcellist, line 112:
```suggestion
      - name: Push Catalog Image
        uses: redhat-actions/push-to-registry@v2.1.7
        with:
          image: eda-server-operator-catalog
          tags: ${{ github.event.inputs.tag_name }} latest
          registry: quay.io/ansible/
          username: ${{ secrets.QUAY_USER }}
          password: ${{ secrets.QUAY_TOKEN }}
```

rooftopcellist:
Thanks for the review!  I made those changes and ran through the workflow again.  This time I also tested the flow of publishing the draft release to trigger the promote workflow, this surfaced a couple more var referencing issues that didn't fail when using the workflow_dispatch flow.  

Everything works now, I am going to go ahead and merge this.  Thanks for the great collaboration here y'all!

goldyfruit:
Awesome :+1:

PR #158, opened 2024-01-02
fix(docker-buildx): use correct buildx variable for cross-compilation

jon-nfc:
The current makefile option `docker-buildx` was not correctly setup for cross-platform compilation.

var `BUILDPLATFORM` is the architecture for the machine building the image. This var is not for cross-compilation.

var `TARGETPLATFORM` is the correct variable to use for cross-compilation.

PR #161, opened 2024-01-08
Enable multi-arch builds in GHA for the main branch

rooftopcellist:
We should build multi-arch images for the eda-server-operator, starting with the the images built for the `main` branch. 

Example of a successful run on my fork:
* https://github.com/rooftopcellist/eda-server-operator/actions/runs/7442607946/job/20246278257

### DIFF
--- a/.github/workflows/build-operator-image.yaml
+++ b/.github/workflows/build-operator-image.yaml
@@ -13,16 +13,24 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
-
-      - name: Login to Quay.io
-        uses: docker/login-action@v1
+      - name: Log into registry ghcr.io
+        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d # v3.0.0
         with:
-          registry: quay.io
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Log into registry quay.io
+        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d # v3.0.0
+        with:
+          registry: quay.io/ansible/
           username: ${{ secrets.QUAY_USER }}
           password: ${{ secrets.QUAY_TOKEN }}
 
-      - name: Build and Push Image
+      - name: Build and Store Image @ghcr
         run: |
-          IMG=quay.io/ansible/eda-server-operator:main make docker-buildx
+          IMG=ghcr.io/${{ github.repository_owner }}/eda-server-operator:${{ github.sha }} make docker-buildx
+
+      - name: Publish Image to quay.io/ansible/eda-server-operator:main
+        run: |
+          docker buildx imagetools create ghcr.io/${{ github.repository_owner }}/eda-server-operator:${{ github.sha }} --tag quay.io/ansible/eda-server-operator:main
--- a/.github/workflows/build-operator-image.yaml
+++ b/.github/workflows/build-operator-image.yaml
@@ -13,15 +13,16 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Build Image
-        run: |
-          IMG=eda-server-operator:main make docker-build
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
 
-      - name: Push To Quay
-        uses: redhat-actions/push-to-registry@v2.1.1
+      - name: Login to Quay.io
+        uses: docker/login-action@v1
         with:
-          image: eda-server-operator
-          tags: main
-          registry: quay.io/ansible/
+          registry: quay.io
           username: ${{ secrets.QUAY_USER }}
           password: ${{ secrets.QUAY_TOKEN }}
+
+      - name: Build and Push Image
+        run: |
+          IMG=quay.io/ansible/eda-server-operator:main make docker-buildx


jon-nfc:
@rooftopcellist, great minds think alike (#162) and if it wasn't for @kurokobo, I would not have noticed your work.

I'll post what I've done so far as it might give some ideas.

.github/workflows/build-operator-image.yaml
```diff
@@ -13,15 +13,16 @@ jobs:
...
-          image: eda-server-operator
-          tags: main
-          registry: quay.io/ansible/
+          registry: quay.io
           username: ${{ secrets.QUAY_USER }}
           password: ${{ secrets.QUAY_TOKEN }}
+
+      - name: Build and Push Image
+        run: |
+          docker buildx build --push --platform=linux/arm64,linux/amd64,linux/s390x,linux/ppc64le --tag quay.io/ansible/eda-server-operator:main -f Dockerfile .
```
jon-nfc, line 28: I took a very different approach to this file the reality is that there was only one line to change, one step to move, and one to add. none of the other steps are required.

``` diff
---

name: Build Operator Image

on:
  push:
    branches: [main]

jobs:
  release:
    runs-on: ubuntu-latest
    name: Push main image
    steps:
      - uses: actions/checkout@v2

+      - name: Log into registry quay.io
+        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d # v3.0.0
+         with:
+           registry: quay.io/ansible/
+           username: ${{ secrets.QUAY_USER }}
+           password: ${{ secrets.QUAY_TOKEN }}

      - name: Build Image
        run: |
-          IMG=eda-server-operator:main make docker-build
+         IMG=quay.io/ansible/eda-server-operator:main make docker-buildx

-      - name: Push To Quay
-        uses: redhat-actions/push-to-registry@v2.1.1
-        with:
-          image: eda-server-operator
-          tags: main
-          registry: quay.io/ansible/
-          username: ${{ secrets.QUAY_USER }}
-          password: ${{ secrets.QUAY_TOKEN }}

```

Docker buildx from the makefile also pushes, so no additional push steps are required. However prior to pushing, you must login.

jon-nfc, line 28: reference #162 - https://github.com/ansible/eda-server-operator/pull/162/files#diff-32245d28b869933b2f68c1b883a196955a5315d1a80b36b645e4a897a522c6e8

The approach I took was to keep the same workflow of the original file, and add an additional step to store the build image on ghcr. This served the purpose of supporting future workflows, releases for instance. as now all that would need to be done for the release workflow is pull the ghcr image and publish it using the git tag.

The release workflow can also be viewed at #162 https://github.com/ansible/eda-server-operator/pull/162/files#diff-e380de8fb76960db94a9cf78262f3034b87215ed566a3d3db167d0b2125399bf.

I've also noticed #159 so will add comments there RE the changes from #158 that are beneficial and what I was doing in #162 .

jon-nfc, line 28: for consistency with comment https://github.com/ansible/eda-server-operator/pull/159#discussion_r1444377284

``` suggestion
      - name: Log into registry ghcr.io
        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d # v3.0.0
        with:
          registry: ghcr.io
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}


      - name: Log into registry quay.io
        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d # v3.0.0
        with:
          registry: quay.io/ansible/
          username: ${{ secrets.QUAY_USER }}
          password: ${{ secrets.QUAY_TOKEN }}


      - name: Build and Store Image @ghcr
        run: |
          IMG=ghcr.io/${{ github.repository_owner }}/eda-server-operator:${{ github.sha }} make docker-buildx


      - name: Publish Image to quay.io/eda-server-operator:main
        run: |
          docker buildx imagetools create ghcr.io/${{ github.repository_owner }}/eda-server-operator:${{ github.sha }} --tag quay.io/ansible/eda-server-operator:main
```

rooftopcellist: That makes me realize, I still had the full custom command in from an intermediary commit when I thought I needed to use `--load`, then  push, rather than just use the `--push` flag.  I just switched back to using the make target.

Thanks for the review!

jon-nfc: thats the beauty of the make file. can be used everywhere you can run a cli command

rooftopcellist: gah, I made a mess by committing another change first, so this commit suggestion wouldn't apply via the UI.  I added this commit and credited it to you in the commit message.  Thanks for this suggestion!
* https://github.com/ansible/eda-server-operator/pull/161/commits/956957d57ea2ca4fc8d5d26fd32feb0b59a31f45

rooftopcellist:
This worked as expected when I tested it on my fork:
* https://github.com/rooftopcellist/eda-server-operator/actions/runs/7450358643/job/20269079697

.github/workflows/build-operator-image.yaml
```diff
@@ -13,15 +13,24 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Build Image
-        run: |
-          IMG=eda-server-operator:main make docker-build
+      - name: Log into registry ghcr.io
+        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d # v3.0.0
```
Alex-Izquierdo, line 17: Just out of curiosity, why use the commit with a comment with the tag instead of just the tag?

Alex-Izquierdo (approved):
LGTM, but why not going far away and be able to generate also a tagged image?

.github/workflows/build-operator-image.yaml
rooftopcellist: @Alex-Izquierdo that change was made by @jon-nfc , but my guess is that it was to pin to a particular known working version of the docker login action in case a new release breaks it.  However, I do not know of a specific issue why this might have been done.  

The commit is correct for v3.0.0:
* https://github.com/docker/login-action/releases/tag/v3.0.0

rooftopcellist:
@Alex-Izquierdo I'm not sure I understood this bit.

> LGTM, but why not going far away and be able to generate also a tagged image?

Currently, it will create an image with the git commit sha as the tag and push it to ghcr.io to be re-tagged for release, or for debugging purposes.  It will also push to quay.io with the `main` tag so that users always have a build of the latest changes in the `main` branch of the operator.

.github/workflows/build-operator-image.yaml
jon-nfc: @Alex-Izquierdo 

@rooftopcellist is correct it originated from myself. Code was copied directly from github action template. However the reason for keeping it this way is very simple, Security.

A word none of us like, but must deal with. Security in this case is the prevention (reduction in probability) of supply chain attack or more correctly put limiting attack surface.

A git hash in the context of a repository will always remain the same and always be in the same location in the git history. You can not change a git hash. Even if you submit the same code in the same location in the git history, the git hash will always be different.

A Git tag is tied to a git hash. The problem with a git tag for chain of custody is that the git tag can move hashes. You can delete an existing tag and attach it to another commit hash. This is problem when using it to reference a dependency.

A hash to reference the same work as a git tag is preferable. one would hope that the developer whom used the code had reviewed the dependent repo. This I suppose you could call now a dependency in a known and  "approved" state. If the hash changes, which would make it not exist, then an error will occur. 

Whilst it's not impossible for a hash collision to occur, the probability of it occurring is in the order of 2^40 which is significantly larger than the tag change which anyone with access to commit can do.

jon-nfc: addendum.

the commented tag version exists next to the hash as we are humans, and readability is important.

PR #167, opened 2024-01-12
Remove unneeded redirect to /eda

rooftopcellist:
This reverts commit 15117a2de9382ae2dcfb4d9c2e6fbf01de067a4e.

Without this change, there are some weird redirect issues when deploying EDA using the operator right now.

If I navigate to:
* https://eda-host/, or 
* https://eda-host/eda, I get page not found, with a 200 (not 404).  However, if I click "Return to Dashboard", it will bring me to the login page, and I can log in and see the dashboard.

image

### DIFF
--- a/roles/eda/templates/eda-ui.deployment.yaml.j2
+++ b/roles/eda/templates/eda-ui.deployment.yaml.j2
@@ -64,13 +64,13 @@ spec:
         - containerPort: 8080
         readinessProbe:
           httpGet:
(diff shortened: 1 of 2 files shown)


rooftopcellist:
Sure enough, in testing, removing the redirect results in a working UI.  Thanks @jamesmarshall24 !

Alex-Izquierdo (approved):

PR #177, opened 2024-01-30
Adding POSTGRES_ variables to support postgres official images

### LINKED ISSUE #175: postgres: Change POSTGRESQL variables names
I was able to bring up postgresql with AWX operator just fine. I also noted that the awx operator support a different set of variables.

Please see:
https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/statefulsets/postgres.yaml.j2

POSTGRESQL_PASSWORD (rhel8 based image) and POSTGRES_PASSWORD (postgres images)

Anyway, the results when bringing up the operator database is:

```
$ kubectl -n eda logs pod/eda-postgres-13-0
Error: Database is uninitialized and superuser password is not specified.
       You must specify POSTGRES_PASSWORD to a non-empty value for the
       superuser. For example, "-e POSTGRES_PASSWORD=password" on "docker run".

       You may also use "POSTGRES_HOST_AUTH_METHOD=trust" to allow all
       connections without a password. This is *not* recommended.

       See PostgreSQL documentation about "trust":
       https://www.postgresql.org/docs/current/auth-trust.html
```

I fixed this easily by simply change the variable name to POSTGRES_ and run the Ansible playbook instead.

While writing this issue, I realized I might have caused this because I had to rewrite the image name to go through a proxy, which I copy and pasted from a working awx operator deployment.

EDA default: quay.io/sclorg/postgresql-13-c9s

AWX default:
postgres_image: postgres
postgres_image_version: 13

I might have assumed that these 2 projects were related. My bad. Something to think about.

smalenfant:
Makes EDA compatible with postgres official images. Block of code taken from awx operator, but modifed for this project.

Reference: https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/statefulsets/postgres.yaml.j2

Closes #175

### DIFF
--- a/roles/postgres/templates/postgres.statefulset.yaml.j2
+++ b/roles/postgres/templates/postgres.statefulset.yaml.j2
@@ -69,6 +69,7 @@ spec:
             initialDelaySeconds: 45
             timeoutSeconds: 2
           env:
+            # For postgres_image based on rhel8/postgresql-13
             - name: POSTGRESQL_DATABASE
               valueFrom:
                 secretKeyRef:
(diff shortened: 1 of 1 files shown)


rooftopcellist (approved):
Thanks for the PR!

PR #180, opened 2024-01-31
Fix Force DB Drop command

rooftopcellist:
Currently, when running the restore role, we see this error:

```
 TASK [Restore database dump to the new postgresql container] ******************************** 
[0;31mfatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'force_drop_db' is undefined\n\nThe error appears to be in '/opt/ansible/roles/restore/tasks/postgres.yml': line 103, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Restore database dump to the new postgresql container\n  ^ here\n"}[0m
```

This is because no default was set in eda/defaults/vars.yml.

This PR resolves that and a few other things to better align with the awx-operator.

### DIFF
--- a/roles/restore/tasks/postgres.yml
+++ b/roles/restore/tasks/postgres.yml
@@ -80,25 +80,34 @@
       -p {{ eda_postgres_port }}
   no_log: "{{ no_log }}"
 
-- name: Set drop db command
-  set_fact:
-    pg_drop_db: >-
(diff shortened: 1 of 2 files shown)


rooftopcellist:
Tested and verified the whole install, backup and restore flow.

PR #181, opened 2024-02-23
Add max running activations default & refresh deployment after configmap change

rooftopcellist:
Add default max running activations value & refresh deployment upon configmap change.

### DIFF
--- a/docs/user-guide/settings.md
+++ b/docs/user-guide/settings.md
@@ -1,0 +1,28 @@
+# EDA Application Settings
+
+
+Here is an example of how to configure settings for the EDA application via the EDA custom resource. This will override the default application settings if set.
+
+```yaml
+---
+apiVersion: eda.ansible.com/v1alpha1
+kind: EDA
+metadata:
+  name: eda
+spec:
+  ...
+  extra_settings:
+    - setting: MAX_RUNNING_ACTIVATIONS
+      value: "12"
+
+```
+
+
+
+## Commonly Customized Settings
+
+Below is a table of the setting name, default value, and a description of it's purpose:
+
+| Setting Name             | Default Value  | Description                                     |
+|--------------------------|----------------|-------------------------------------------------|
+| MAX_RUNNING_ACTIVATIONS  | "12"           | Maximum number of running activations           |
--- a/README.md
+++ b/README.md
@@ -214,6 +214,7 @@ spec:
 
 ### Additional Advanced Configuration
 - [No Log](./docs/user-guide/advanced-configuration/no-log.md)
+- [EDA application settings](./docs/user-guide/advanced-configuration/settings.md)
 - [Deploy a Specific Version of EDA](./docs/user-guide/advanced-configuration/deploying-a-specific-version.md)
 - [Trusting a Custom Certificate Authority](./docs/user-guide/advanced-configuration/trusting-a-custom-certificate-authority.md)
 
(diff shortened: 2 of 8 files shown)


rooftopcellist:
In testing, on a fresh deploy, I see the setting being set:


```
$ oc rsh eda-demo-api-574d5b697f-lkx6z
sh-5.1$ env | grep MAX
EDA_MAX_RUNNING_ACTIVATIONS=12
```

Then, when I change the EDA custom resource spec to:

```yaml

kind: EDA
metadata:
  name: eda-demo
spec:
  extra_settings:
    - setting: EDA_MAX_RUNNING_ACTIVATIONS
      value: 11

```

I see the reconciliation loop gets triggered, and the configmap gets correctly updated:

```
$ oc get cm -o yaml eda-demo-eda-env-properties
apiVersion: v1
data:
  EDA_ALLOWED_HOSTS: '[''*'']'
  EDA_CONTROLLER_SSL_VERIFY: "no"
  EDA_CONTROLLER_URL: https://awx-awx.apps.aap-dev.ocp4.testing.ansible.com
  EDA_CSRF_TRUSTED_ORIGINS: http://eda-demo-api:8000
  EDA_DEPLOYMENT_TYPE: k8s
  EDA_HOST: eda-demo-api:8000
  EDA_MAX_RUNNING_ACTIVATIONS: "11"
  EDA_MQ_HOST: eda-demo-redis-svc
  EDA_PROTOCOL: http
  EDA_WEBSOCKET_BASE_URL: ws://eda-demo-daphne:8001
  EDA_WEBSOCKET_SSL_VERIFY: "False"

```

Finally, I see that the api pod gets cycled to pick up the new value. After it comes up:

```
$ oc rsh eda-demo-api-746f6b8c7d-qh8fv
sh-5.1$ env | grep MAX
EDA_MAX_RUNNING_ACTIVATIONS=11
```

docs/user-guide/settings.md
```diff
@@ -0,0 +1,30 @@
...
+kind: EDA
+metadata:
+  name: eda
+spec:
+  ...
+  extra_settings:
+    - setting: MAX_RUNNING_ACTIVATIONS
+      value: "12"
+    - setting: GIT_SSL_NO_VERIFY
+      value: "true"
```
kurokobo, line 18:
```suggestion
```
AFAIK, passing `GIT_SSL_NO_VERIFY` through environment variables does not take any effect. Any environment variables for worker pod will not be inherited for `git` command, since `subprocess.run()` is invoked with explicit `env` param with minial dict: https://github.com/ansible/eda-server/blob/7e956b8810987fce20de698ad05a7f3e3acc323f/src/aap_eda/services/project/git.py#L192

So this two lines may cause users to be confusing. I want to suggest you to remove these lines.

docs/user-guide/settings.md
```diff
@@ -0,0 +1,30 @@
...
+
+
+## Commonly Customized Settings
+
+Below is a table of the setting name, default value, and a description of it's purpose:
+
+| Setting Name             | Default Value  | Description                                     |
+|--------------------------|----------------|-------------------------------------------------|
+| MAX_RUNNING_ACTIVATIONS  | "12"           | Maximum number of running activations           |
+| GIT_SSL_NO_VERIFY        | "true"         | Disable SSL verification for Git operations     |
```
kurokobo, line 30:
```suggestion
```

As the same as above comment.

kurokobo (commented):
Hi, I want to leave a minor comment. Thanks!

docs/user-guide/settings.md
rooftopcellist: Thank you @kurokobo ! I'll remove it from the examples.

aknochow (approved):

Issue #138, opened 2023-10-19
Unable to launch managed postgres db in cluster


shrutebattlestargalactica:
Hello @rooftopcellist! I am having some issues with deploying EDA into our EKS cluster:

[eda-manager.log](https://github.com/ansible/eda-server-operator/files/13047476/eda-manager.log)

it is stuck on 

```
TASK [postgres : Wait for Database to initialize if managed DB]:

{"level":"info","ts":1697747113.740688,"logger":"proxy","msg":"cache miss: /v1, Kind=PodList err-Index with name field:status.phase does not exist"}
```

I deployed via recommended operator with ArgoCD/Customize

Are we not able to yet use an external db similar to what we do with AWX? thanks for your help!

shrutebattlestargalactica:
@rooftopcellist I am wonder if this is related to this issue similar to the awx-operator: https://github.com/ansible/awx-operator/issues/1022

I forgot to mention but we run this EKS cluster on k8s v1.28

shrutebattlestargalactica:
Thats correct - the pod isnt crashing due to this error I posted:

{"level":"info","ts":1697747113.740688,"logger":"proxy","msg":"cache miss: /v1, Kind=PodList err-Index with name field:status.phase does not exist"}


I have a very vanilla set up for the deployment:

`# eda.yaml
apiVersion: eda.ansible.com/v1alpha1
kind: EDA
metadata:
  name: awxnonprod-eda
  namespace: eda
spec:
  automation_server_url: https://awxdev.mydomain.com
  service_type: nodeport
  ingress_type: none
  service_labels: |
    environment: nonprod
    team: technology
  hostname: edadev.mydomain.com`

shrutebattlestargalactica:
Alright I ended up getting the external RDS postgres working! I was mimicking my secret from the AWX configuration and had to remove the "type: unmanaged" from the secret below:

> ---
> apiVersion: v1
> kind: Secret
> metadata:
>   name: eda-nonprod-postgres-configuration
>   namespace: eda-nonprod
> stringData:
>   host: "mydatabase.rds.amazonaws.com"
>   port: "5432"
>   database: "edadbnonprod"
>   username: "edaadmin"
>   password: "supersecretpassword"
>   sslmode: "prefer"
> type: Opaque

shrutebattlestargalactica:
[...]

PR #190, opened 2024-04-05
add postgres_security_context_settings

rchaud:
Adding the option to add custom security context for Postgres as in K8s it does not start due to a permission issue. This allows to change the user id, group id, etc to anything other than 26 used by the image https://quay.io/repository/sclorg/postgresql-13-c9s

https://github.com/ansible/eda-server-operator/issues/138

Example usage:
```
postgres_security_context_settings:
    runAsUser: 1001
    fsGroup: 1001
```

### DIFF
--- a/roles/postgres/templates/postgres.statefulset.yaml.j2
+++ b/roles/postgres/templates/postgres.statefulset.yaml.j2
@@ -31,6 +31,9 @@ spec:
         app.kubernetes.io/managed-by: 'eda-operator' 
     spec:
       securityContext:
+{% if postgres_security_context_settings|length %}
+        {{ postgres_security_context_settings | to_nice_yaml | indent(8) }}
+{% endif %}
         allowPrivilegeEscalation: false
         capabilities:
           drop: ["ALL"]


rooftopcellist:
@chinochao Thanks for the PR!

Ultimately, we intend to nest parameters like these under `database`.  We want to do this nested approach with all of the operators eventually, but for the awx-operator for example, it will require us to create a new apiVersion to do so.

With eda-server-operator, we have the luxury of a fresh start and can do this now.  Would you be open to re-working the PR?

The user could then define it on the EDA spec like this:

```
spec:
  database:
    security_context:
      runAsUser: 1001
      fsGroup: 1001
```

We'll also need CRD and CSV entries, like these, but nested under `database`:
* https://github.com/ansible/awx-operator/blob/7b02b5df04ba7221ea09666d06120f9dc02893c1/config/crd/bases/awx.ansible.com_awxs.yaml#L1887
* https://github.com/ansible/awx-operator/blob/7b02b5df04ba7221ea09666d06120f9dc02893c1/config/manifests/bases/awx-operator.clusterserviceversion.yaml#L113

rooftopcellist:
@chinochao I was trying to reproduce the actual bug first while testing this out but could not while using k3s locally.  Could you share more details about your deployment?  It'd be good to know under what circumstances this permissions error might occur, so that we can document the fix better.

maratsal:
you can reproduce this with deployment on EKS with gp2 storage class (based on aws-ebs)
```
kubectl get storageclass
NAME            PROVISIONER             RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
gp2 (default)   kubernetes.io/aws-ebs   Delete          WaitForFirstConsumer   false                  390d
```

kaiokmo:
housekeeping note: the current code on `main` already has a pod-level securityContext on the postgres StatefulSet, and there's a separate container-level securityContext on the init container

I'm closing this PR

PR #197, opened 2024-05-10
eda: Add WEBSOCKET_TOKEN_BASE_URL to settings

dsavineau:
The activation workers need to access to the api (gunicorn) for refreshing the token.

This is done via the `WEBSOCKET_BASE_URL` value but both gunicorn and daphne need to be available via this url.

However for k8s deployment `WEBSOCKET_BASE_URL` points directly to the daphne service rather than going back to nginx.

https://github.com/ansible/eda-server/commit/3353129


Alex-Izquierdo (approved):
LGTM but should this be aligned with https://github.com/ansible/eda-server-operator/pull/198 ?

rooftopcellist (approved):

dsavineau:
> LGTM but should this be aligned with #198 ?

I'll rebase #198 once this one is merged

PR #206, opened 2024-06-03
Add default replicas entries back to the CRD

rooftopcellist:
- Fixes a bug where scheduler replicas were getting set to 0 when creating an EDA CR via the UI form because of the minimum: 0 field

### DIFF
--- a/config/crd/bases/eda.ansible.com_edas.yaml
+++ b/config/crd/bases/eda.ansible.com_edas.yaml
@@ -92,6 +92,7 @@ spec:
                     description: 'The number of gunicorn workers for the api.
                       Default: 2'
                     format: int32
+                    default: 2
                     minimum: 1
                     nullable: false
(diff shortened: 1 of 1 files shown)


config/crd/bases/eda.ansible.com_edas.yaml
```diff
@@ -104,7 +105,8 @@ spec:
                     description: 'Size is the size of number of eda-api replicas.
                       Default: 1'
                     format: int32
-                    minimum: 0
+                    default: 1
```
Alex-Izquierdo, line 108: Should be the default "2" as we already do with the other workers?

rooftopcellist: talked out of band and default of 1 is OK for now for API.  We can document setting it to 2 for redundancy if desired.

PR #207, opened 2024-06-03
Add wait-for-migrations initContainer to scheduler and worker deployents

rooftopcellist:
Add wait-for-migrations initContainer to scheduler and worker deployments. 

The configmap approach won't work because of permissions issues on Openshift:

```
$ oc logs -f -c wait-for-migrations eda-scheduler-9cdbdf445-cb4mc 
chmod: changing permissions of '/usr/bin/wait-for-migrations': Read-only file system
```

So we will need to add the script to the eda-server container.  
* https://github.com/ansible/eda-server/pull/925


roles/eda/templates/eda-activation-worker.deployment.yaml.j2
```diff
@@ -66,6 +66,54 @@ spec:
 {% endif %}
 {% if bundle_ca_crt  %}
       initContainers:
+      - name: wait-for-migrations
+        image: {{ _image }}
+        imagePullPolicy: '{{ image_pull_policy }}'
+        command:
+          - /bin/bash
```
Alex-Izquierdo, line 73: you can run `/bin/bash /usr/bin/wait-for-migrations` so you don't need to run chmod +x

Alex-Izquierdo: and/or, you can mount the file in a place where you can have write perms

rooftopcellist: I tried those suggestions and a couple other ideas, no luck.  But baking in the script into the image worked.

kurokobo: I don't have any OpenShift but I think we can `defaultMode` for the configmap for the `volumes` to change perms for the script file initially.

```yaml
  volumes:
    ...
    - name: wait-for-migrations-script
      configMap:
        name: wait-for-migrations-script-configmap
        defaultMode: 0500
        items:
          - key: wait-for-migrations
            path: wait-for-migrations
```

Refer to: https://access.redhat.com/solutions/7024882

rooftopcellist: Oh! Thanks kurokobo!  I will make use of that pattern in the future for sure.  After talking with @Alex-Izquierdo , I think we are going to go with the management command approach implemented here:
* https://github.com/ansible/eda-server/pull/927/files

It will be a nice way to have the "script" available in all installation methods for EDA.

Alex-Izquierdo: @rooftopcellist the PR is merged.

roles/eda/templates/eda-activation-worker.deployment.yaml.j2
```diff
@@ -66,6 +66,50 @@ spec:
 {% endif %}
 {% if bundle_ca_crt  %}
       initContainers:
+      - name: wait-for-migrations
```
dsavineau, line 69: I think we need to move this (including the `initContainers` statement) outside of the bundle_ca_crt condition otherwise this won't work on default deployment

rooftopcellist: Good spot, I just pushed a change to move it outside of the `bundle_ca_crt` block.  Thanks!

dsavineau (approved):

aknochow (approved):

PR #209, opened 2024-06-10
Ignore errors when creating Superuser, look for specific output instead

rooftopcellist:
There is a temporary error that is hit when deploying EDA when this task runs and migrations have not yet been completed.  

```
 TASK [Create super user via Django if it doesn't exist.] ******************************** 
fatal: [localhost]: FAILED! => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result"}
```

We should ignore errors to keep the logs clean. Subsequent reconciliation loops will handle the user creation.


rooftopcellist:
Closing.  We should instead have a task that waits for the migrations to be complete.

PR #210, opened 2024-06-12
Add ability to disable Redis

rooftopcellist:
This PR makes it possible to disable redis by setting `eda.spec.redis.disabled: true`.  This might be desired if configuring an external redis instance via the env settings.  

```
  extra_settings:
    - setting: EDA_MQ_USER
      value: '{{ eda_redis_username | b64decode }}'
    - setting: EDA_MQ_USER_PASSWORD
      value: '{{ eda_redis_password | b64decode }}'
    - setting: EDA_MQ_HOST
      value: '{{ eda_redis_host | b64decode }}'
    - setting: EDA_MQ_PORT
      value: '{{ eda_redis_port | b64decode }}'
```

**Alternative Idea:**

However, I think it might be better all around to make it so that the redis configuration is done via a k8s secret, and the presence of that k8s secret being specified on the EDA spec would indicate that an external Redis instance is being used configured and that the default redis pod/deployment/service should not be created.  

@rcarrillocruz What do you think about that idea?  If that's better, I can close this and implement it tomorrow.

### DIFF
--- a/config/crd/bases/eda.ansible.com_edas.yaml
+++ b/config/crd/bases/eda.ansible.com_edas.yaml
@@ -2076,6 +2076,9 @@ spec:
               redis:
                 description: 'Defines desired state of cache resources'
                 properties:
+                  disabled:
+                    description: Disables default redis deployment if external redis is configured
+                    type: boolean
                   replicas:
                     description: 'Defines Redis Deployment replicas'
                     format: int32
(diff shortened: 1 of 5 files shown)


rooftopcellist:
Discussed oob.  I'll close this PR and open one to make it configurable via a redis secret that can be swapped out for External Redis Configuration secret.  And then we'll key off of it's presence to determine whether or not to create the default redis deployment/svc/pod etc.

PR #214, opened 2024-07-04
node_selector and tolerations docs

dmarav:
Updated documentation for `node_selector` and `tolerations` changes made in #205 .


rooftopcellist (approved):
Thank you for adding these docs @dmarav

PR #216, opened 2024-07-16
Add an nginx image to the api deployment to serve up api static files

rooftopcellist:


roles/eda/templates/eda-api.configmap.yaml.j2
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        include mime.types;
+        types {
+            application/manifest+json webmanifest;
+        }
+        server {
+            listen {{ api_nginx_port }};
+            location ^~ /static/ {
+                alias {{ static_path }}/;
```
Zokormazo, line 24: I did not hack around nginx conf for a while, but I was expecting to find a `root` statement here pointing to the filesystem location of the static assets

roles/eda/vars/main.yml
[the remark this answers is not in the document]
rooftopcellist: Agreed on calling it `api_django_port`. I just made that change.

I'd like to handle renaming "internal variables" to underscore variables for now if that's OK.  I'd like to talk more on it an come up with a _standard_, then do it everywhere.  It would be good to know if there is an ansible standard already out there.  Seems like having the variables set in the vars/main.yml already marks them as "variables not to be changed by the user" (those are in defaults/main.yml).  I've typically tried to name set_fact task variables with a preceding underscore.  Though, I have not been consistent across the board...

benthomasson (approved):

roles/eda/templates/eda-api.configmap.yaml.j2
rooftopcellist: The root directive is not strictly necessary in this configuration because the alias directive is being used in the `/static/` location block. The alias directive allows Nginx to map requests for `/static/` to a different directory on the filesystem, specified by `{{ static_path }}/`.

So I think we're good.

rooftopcellist:
The changes in scope for this PR work.  There is a separate problem we found while testing where projects do not sync/clone.  We will resolve that in a follow-up PR to eda-server, and the operator as needed.

PR #217, opened 2024-07-17
Set ANSIBLE_LOCAL_TEMP as /tmp for project clones to work

rooftopcellist:


benthomasson:
This is probably not needed.   We need to update the environment in the code itself since it doesn't inherit from the container environment.

PR #219, opened 2024-07-19
(WIP) Fix: Do not cause operator failure when 'bundle_cacert_secret' is incorrectly specified

zkayyali812:
This PR is attempts to add the logic to avoid failures when a `bundle_cacert_secret` is incorrectly specified in the PR.

Wondering if we should also add a `Condition` to the Status of our EDA resource to specify this? And if so, do we have any existing patterns for this?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ kustomization.yaml
 *.swo
 *~
 .vscode/
+venv/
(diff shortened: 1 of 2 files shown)


kaiokmo:
housekeeping note: I'm closing this one as stale, and also because this is not a bug in the operator itself.

the scenario this PR is guarding against seems to be: the user specifies `bundle_cacert_secret: my-custom-certs` in the CR, but the secret `my-custom-certs` doesn't actually exist in the namespace

that's a user misconfiguration, and arguably the operator should fail loudly rather than silently swallowing the error. a missing CA secret means TLS verification against external services will break anyway. silently continuing just delays the failure to a more confusing point and would make debugging harder.

PR #221, opened 2024-07-22
Rework URL status task to look for the API or the UI route

rooftopcellist:
Rework URL status task to look for the API or the UI route
- If the UI route does not exist, the API route (appended -api) should be used.


rooftopcellist:
Oops, I see @dsavineau 's implementation now.  That's better because it is explicit.  Let's go with his approach there. 
* https://github.com/ansible/eda-server-operator/pull/220

PR #224, opened 2024-07-24
Make /api/eda and /api/eda/v1 endpoints available via API service

rooftopcellist:
@Alex-Izquierdo this is not yet working, could I get your feedback on where I've gone wrong?

### DIFF
--- a/roles/eda/templates/eda.configmap.yaml.j2
+++ b/roles/eda/templates/eda.configmap.yaml.j2
@@ -24,6 +24,8 @@ data:
   EDA_WEBSOCKET_SSL_VERIFY: "{{ websocket_ssl_verify }}"
   EDA_MAX_RUNNING_ACTIVATIONS: "12"
 
+  EDA_STATIC_URL: /api/eda/static/
+
   # Custom user variables
 {% for item in extra_settings | default([]) %}
   {{ item.setting | upper }}: "{{ item.value }}"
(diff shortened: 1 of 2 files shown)


Alex-Izquierdo (approved):
I think it is fine as it is.

Alex-Izquierdo:
As we discussed privately, the nginx conf of eda-ui needs to be changed as well.

roles/eda/templates/eda.configmap.yaml.j2
```diff
@@ -75,21 +77,19 @@ data:
...
                 proxy_set_header Origin http://{{ api_server_name }}:{{ api_nginx_port }};
                 proxy_set_header Host $http_host;
-                proxy_set_header X-Forwarded-Host $host;
                 proxy_set_header X-Forwarded-Proto $remote_scheme;
                 proxy_set_header X-Forwarded-Port $server_port;
                 proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
             }
 
-            location ~ ^/static/ {
+            location ~ ^/api/eda/static/ {
```
Alex-Izquierdo, line 89: you don't need this location, is mostly duplicated with the previous one. `location ~ ^/api/eda`

rooftopcellist:
Updated.  Also, now redirects work with curls for /api/eda when using the api service directly.

```
sh-5.1$ curl eda-api:8000/api/eda -k -L -u admin:password
{"current_version":"http://eda-api:8000/api/eda/v1/","available_versions":{"v1":"http://eda-api:8000/api/eda/v1/"}}
```

```
sh-5.1$ curl eda-ui:80/api/eda -k -L -u admin:password
{"current_version":"http://eda-ui/api/eda/v1/","available_versions":{"v1":"http://eda-ui/api/eda/v1/"}}
```

PR #227, opened 2024-07-31
Move bundle_ca_crt conditional to not include static files

rooftopcellist:
Prior to this change, if the bundle ca secret was not configured, the api endpoints would not be rendered.

This is because the collectstatic management command would be a noop because the static directory was not created by the volumeMount.

### DIFF
--- a/roles/eda/templates/eda-api.deployment.yaml.j2
+++ b/roles/eda/templates/eda-api.deployment.yaml.j2
@@ -153,10 +153,10 @@ spec:
 {% if combined_api.resource_requirements is defined %}
         resources: {{ combined_api.resource_requirements }}
 {% endif %}
-{% if bundle_ca_crt  %}
         volumeMounts:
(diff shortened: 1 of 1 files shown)


rooftopcellist:
I tested this out when not configuring bundle ca cert, and it now works.

```
$ oc logs -f myaap-eda-api-6dddd5b5bb-d8ppk -c eda-initial-data
Added role "Admin" with 34 permissions to 10 resources
Added role "Organization Member" with 2 permissions to 1 resources
Added role "Editor" with 21 permissions to 10 resources
Added role "Contributor" with 23 permissions to 9 resources
Added role "Operator" with 12 permissions to 9 resources
Added role "Auditor" with 10 permissions to 10 resources
Added role "Viewer" with 10 permissions to 10 resources
Added 7 roles.

36 static files copied to '/var/lib/ansible-automation-platform/eda/static'.
```

PR #228, opened 2024-08-05
Initial webhooks deployment feature

rooftopcellist:
cc @msmagnanijr

### DIFF
--- a/roles/eda/templates/eda-webhook.deployment.yaml.j2
+++ b/roles/eda/templates/eda-webhook.deployment.yaml.j2
@@ -2,40 +2,40 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: "{{ ansible_operator_meta.name }}-activation-worker"
-  namespace: "{{ ansible_operator_meta.namespace }}"
   labels:
     {{ lookup("template", "../common/templates/labels/common.yaml.j2")  | indent(width=4) | trim }}
     {{ lookup("template", "../common/templates/labels/version.yaml.j2") | indent(width=4) | trim }}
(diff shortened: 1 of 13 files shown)


msmagnanijr:
@mkanoor could you please take a look? tks

msmagnanijr (approved):
From what we can see it looks okay. Let's see if someone else can review this to make sure.

roles/eda/templates/eda-webhook.deployment.yaml.j2
```diff
@@ -0,0 +1,235 @@
...
+        imagePullPolicy: '{{ image_pull_policy }}'
+        command:
+          - /bin/bash
+          - -c
+          - aap-eda-manage wait_for_migrations -t 60
+        envFrom:
+          - configMapRef:
+              name: '{{ ansible_operator_meta.name }}-{{ deployment_type }}-env-properties'
+        env:
+        - name: EDA_DB_HOST
```
mkanoor, line 79: @rooftopcellist we need EDA_ACTIVATION_DB_HOST here which should be the same ad EDA_DB_HOST

rooftopcellist: @mkanoor is this needed in other deplooyments? (default-worker and activation-worker)?  Or just this webhook one?

Also, does the webhook database instance share the pg credentials? or do we need to generate a separate PostgreSQL connect k8s secret?

Alex-Izquierdo: @rooftopcellist it is required for all, but I see you already added it to the them.

PR #230, opened 2024-08-06
Add Parameter for EDA Webhook Base URL in Configuration Files

msmagnanijr:
This PR adds a parameter to config/crd/bases/eda and templates this into the EDA settings configmap as EDA_WEBHOOK_BASE_URL.

### DIFF
--- a/roles/eda/templates/eda.configmap.yaml.j2
+++ b/roles/eda/templates/eda.configmap.yaml.j2
@@ -12,7 +12,7 @@ data:
   EDA_DEPLOYMENT_TYPE: "k8s"
 
   # Public URL
-  EDA_WEBHOOK_BASE_URL: "{{ public_base_url }}"
+  EDA_WEBHOOK_BASE_URL: "{{ public_base_url.rstrip('/') }}/{{ eda_webhook_prefix_path.lstrip('/') }}"
 
   # EDA Server
   EDA_CONTROLLER_URL: "{{ automation_server_url }}"
--- a/roles/eda/templates/eda.configmap.yaml.j2
+++ b/roles/eda/templates/eda.configmap.yaml.j2
@@ -11,6 +11,9 @@ data:
   # Operator specific settings
   EDA_DEPLOYMENT_TYPE: "k8s"
 
+  # Public URL
+  EDA_WEBHOOK_BASE_URL: "{{ public_base_url }}"
+
   # EDA Server
   EDA_CONTROLLER_URL: "{{ automation_server_url }}"
   EDA_CONTROLLER_SSL_VERIFY: "{{ automation_server_ssl_verify | default('yes')}}"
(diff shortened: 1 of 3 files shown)


roles/eda/templates/eda.configmap.yaml.j2
```diff
@@ -11,6 +11,9 @@ data:
   # Operator specific settings
   EDA_DEPLOYMENT_TYPE: "k8s"
 
+  # Public Hostname
+  EDA_WEBHOOK_BASE_URL: "{{ public_hostname }}"
```
Alex-Izquierdo, line 15: @msmagnanijr this must be "http[s]://{{ public_hostname }}{{ eda_webhook_prefix_path }}"

@rooftopcellist I wonder if it can be `public_base_url` so it can include the scheme and we don't have to figure out it nor assume it is https.

Alex-Izquierdo (changes requested):

roles/eda/templates/eda.configmap.yaml.j2
```diff
@@ -11,6 +11,9 @@ data:
   # Operator specific settings
   EDA_DEPLOYMENT_TYPE: "k8s"
 
+  # Public Hostname
+  EDA_WEBHOOK_BASE_URL: http[s]://{{ public_hostname }}{{ eda_webhook_prefix_path }}
```
Alex-Izquierdo, line 15: @msmagnanijr When I wrote `http[s]` I meant that it can be `http` or `https` that's why I suggest to define `public_base_url` instead of `public_hostname`

Alex-Izquierdo (changes requested):

roles/eda/templates/eda.configmap.yaml.j2
msmagnanijr: ahh got it! I thought that somehow this was already defined. let me update that.

rooftopcellist (approved):

roles/eda/templates/eda.configmap.yaml.j2
```diff
@@ -11,6 +11,9 @@ data:
   # Operator specific settings
   EDA_DEPLOYMENT_TYPE: "k8s"
 
+  # Public URL
+  EDA_WEBHOOK_BASE_URL: "{{ public_base_url }}"
```
Alex-Izquierdo, line 15: This is wrong. As [I commented in my first review](https://github.com/ansible/eda-server-operator/pull/230#discussion_r1705991192), it must contain the prefix_path that is configured for eda. 
Please @msmagnanijr can you just use this code? 

```suggestion
  EDA_WEBHOOK_BASE_URL: "{{ public_base_url.rstrip('/') }}/{{ eda_webhook_prefix_path.lstrip('/') }}"
```

Alex-Izquierdo (changes requested):

Alex-Izquierdo (approved):

rooftopcellist (approved):
Thanks @msmagnanijr !

PR #231, opened 2024-08-07
Make default webhook host different from ui webhook host if ingress i…

rooftopcellist:
Make default webhook host different from ui webhook host if ingress is used.

CI was failing because the webhook admission validator fails on certain clusters if the host of an ingress is the same as the host value of another ingress.  Now that we have 2 ingress (ui and webhook), that is a problem.  

This also makes the webhook ingress host customization.

### DIFF
--- a/roles/eda/templates/eda-ui.ingress.yaml.j2
+++ b/roles/eda/templates/eda-ui.ingress.yaml.j2
@@ -27,6 +27,13 @@ spec:
                 name: '{{ ansible_operator_meta.name }}-ui'
                 port:
                   number: 80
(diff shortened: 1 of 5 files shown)


roles/eda/templates/eda-ui.ingress.yaml.j2
```diff
@@ -27,6 +27,13 @@ spec:
                 name: '{{ ansible_operator_meta.name }}-ui'
                 port:
                   number: 80
+          - path: '/eda_webhooks'
```
Alex-Izquierdo, line 30: I think this should be as well {{ eda_webhook_prefix_path }}

PR #232, opened 2024-08-07
Ci revert webhooks test [DO NOT MERGE]

rooftopcellist:
Checking to see if the webhook changes are what caused CI to start failing (even though they passed on the webhooks PR....)


rooftopcellist:
The error I thought was related to the eda-server image turned out to be flake..

PR #237, opened 2024-08-14
Fix webhook prefix configuration

rooftopcellist:
Ironing out configuration issues.  

Still need to figure out how to auto-set EDA_WEBHOOK url style env vars for ingress... 
Right now it's configurable by public_base_url and webhooks.prefix


msmagnanijr (approved):
lgtm

PR #240, opened 2024-08-16
make webhook path configurable

drodowic:


matburt (dismissed):

kaiokmo:
housekeeping note: webhook ingress no longer exists on `main`. it was removed when event streams replaced the old webhook mechanism. since this is a dead feature now, I'm closing this PR.

PR #241, opened 2024-08-19
Switch to using a secret for Redis properties

dhageman:
The operator currently configures the Redis properties including username/password in a configmap in plain text.

This pull request switches to using a Kubernetes Secret to align with how the other platform operators configure Redis for operations.   The operator is extended to easily support external Redis instances by passing in the name of the secret value to use for configuring Redis operations.

PR #245, opened 2024-08-29
Expose MQ_TLS setting for activating TLS

dhageman:
This will expose the new MQ_TLS setting to force the use of TLS on Redis connections.

This requires this PR to be merged:  https://github.com/ansible/eda-server/pull/1033

### DIFF
--- a/roles/eda/templates/eda.configmap.yaml.j2
+++ b/roles/eda/templates/eda.configmap.yaml.j2
@@ -26,7 +26,6 @@ data:
   # EDA Server
   EDA_ALLOWED_HOSTS: "['*']"
   EDA_CSRF_TRUSTED_ORIGINS: "http://{{ api_server_name }}:{{ api_nginx_port }}"
-  EDA_MQ_HOST: "{{ ansible_operator_meta.name }}-redis-svc"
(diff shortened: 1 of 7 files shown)


roles/eda/templates/eda.configmap.yaml.j2
```diff
@@ -26,7 +26,6 @@ data:
   # EDA Server
   EDA_ALLOWED_HOSTS: "['*']"
   EDA_CSRF_TRUSTED_ORIGINS: "http://{{ api_server_name }}:{{ api_nginx_port }}"
-  EDA_MQ_HOST: "{{ ansible_operator_meta.name }}-redis-svc"
```
Alex-Izquierdo, line 29: Why this line is removed? I think it is a mistake,

dhageman: No mistake!  It is now being set in the Redis configuration secret.  I should have removed it in https://github.com/ansible/eda-server-operator/pull/241 but I missed it.

Alex-Izquierdo (approved):
LGTM @rooftopcellist take a look just in case if you can.

rooftopcellist (approved):
Thanks @dhageman and @Alex-Izquierdo !!!

PR #251, opened 2024-09-23
OCP Redis failover workaround

zkayyali812:
When Redis is running in cluster mode, if a given EDA worker pod is detected with `crashloopbackoff` status, trigger the restart of all EDA worker pods.

### DIFF
--- a/roles/redis/tasks/ensure_eda_workers_redis_connectivity.yml
+++ b/roles/redis/tasks/ensure_eda_workers_redis_connectivity.yml
@@ -14,14 +14,17 @@
     label_selectors:
       - "app.kubernetes.io/component={{ item }}"
   loop: "{{ _eda_workers_deployment_list }}"
+  no_log: "{{ no_log }}"
   register: _eda_worker_pod_info
 
 - name: "Check if restart is required"
   vars:
-    _eda_worker_container_status: "{{ item.resources[0].status.containerStatuses[0].state.waiting.reason | default('Running') }}"
+    _eda_worker_crashloopback_check: "{{ item.resources[0].status.containerStatuses[0].state.waiting.reason | default('Running') }}"
+    _eda_worker_completed_check: "{{ item.resources[0].status.containerStatuses[0].lastState.terminated.reason | default('Running') }}"
   ansible.builtin.set_fact:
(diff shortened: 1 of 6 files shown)


roles/redis/tasks/ensure_eda_workers_redis_connectivity.yml
```diff
@@ -0,0 +1,39 @@
...
+  no_log: "{{ no_log }}"
+  register: _eda_worker_pod_info
+
+- name: "Check if restart is required"
+  vars:
+    _eda_worker_crashloopback_check: "{{ item.resources[0].status.containerStatuses[0].state.waiting.reason | default('Running') }}"
+    _eda_worker_completed_check: "{{ item.resources[0].status.containerStatuses[0].lastState.terminated.reason | default('Running') }}"
+  ansible.builtin.set_fact:
+    _eda_workers_restart_required: true
+  when: _eda_worker_crashloopback_check == 'CrashLoopBackOff' or _eda_worker_completed_check == 'Completed'
```
Alex-Izquierdo, line 26: Is this enough to detect the scenario? What if the worker has been restarted successfully?

roles/redis/templates/redis.secret.yaml.j2
```diff
@@ -17,3 +17,4 @@ stringData:
   port: '6379'
   host: {{ ansible_operator_meta.name }}-redis-svc
   type: 'managed'
+  redis_mode: 'standalone'
```
Alex-Izquierdo, line 20: This would require to handle a new variable, but we already know if it is a cluster with the current instrumentation based on EDA_MQ_REDIS_HA_CLUSTER_HOSTS (cluster_endpoint) What do you think?

zkayyali812: Hmmm. I can look into that implementation of `EDA_MQ_REDIS_HA_CLUSTER_HOSTS` to check. I did see that when redis is installed in `cluster` mode, that this secret does have a `redis_mode` value of `cluster`. 

However our default is standalone. I figured initially that setting it to `standalone` here made it more consistent. However, I can look into using this variable instead to check if Redis is in cluster/standalone mode if that is recommended here.

PR #255, opened 2024-10-04
WIP: Allow PostGres sslmode and sslrootcert

zkayyali812:
Work in progress.

The goal of this PR is to allow for both sslmode and sslrootcert to be specified as part of the PostGres connection


mkanoor:
@zkayyali812 Once this is enabled SER will stop working. Since we currently don't have a way to deliver the Certificates to Activation.

PR #260, opened 2024-11-01
Fix Redirect Page logo and favicon path

rooftopcellist:
Because static files are served up at `/api/eda/static` in the nginx conf, some of the paths for the logo and favicon on the redirect page need to be updated.

### DIFF
--- a/roles/eda/templates/redirect-page.configmap.html.j2
+++ b/roles/eda/templates/redirect-page.configmap.html.j2
@@ -14,11 +14,11 @@ data:
         <title>Redirecting to Ansible Automation Platform</title>
 
         <!-- Favicon links -->
-        <link rel="icon" type="image/x-icon" href="static/rest_framework/docs/img/favicon.ico">
+        <link rel="icon" type="image/x-icon" href="/api/eda/static/media/favicon.ico">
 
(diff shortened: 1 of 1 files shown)


roles/eda/templates/redirect-page.configmap.html.j2
```diff
@@ -14,11 +14,11 @@ data:
         <title>Redirecting to Ansible Automation Platform</title>
 
         <!-- Favicon links -->
-        <link rel="icon" type="image/x-icon" href="static/rest_framework/docs/img/favicon.ico">
+        <link rel="icon" type="image/x-icon" href="/api/eda//static/rest_framework/docs/img/favicon.ico">
```
zkayyali812, line 17:
```suggestion
        <link rel="icon" type="image/x-icon" href="/api/eda/static/rest_framework/docs/img/favicon.ico">
```
Is the // here correct? Or was this unintentional?

rooftopcellist: Thanks for the review!  That was a bad replace, it is fixed now.  I also changed the favicon path to the new path.  I'll DM you with the other PR for that change in the container.

zkayyali812 (approved):

PR #266, opened 2025-01-09
Reorder nginx container resource requirements in template to fix syntax

rooftopcellist:
The api-deployment template does not apply successfully because the resource requirements are being templated in outside of the nginx block in the deployment yaml.  It just needs to be moved up a line and it will work.


rcarrillocruz (approved):

dsavineau (approved):

dsavineau:
unrelated CI failure

PR #269, opened 2025-01-16
AAP-38746 Increase limits in manager.py

lahavyuv86:
Closes https://issues.redhat.com/browse/AAP-38746


config/manager/manager.yaml
```diff
@@ -97,8 +97,8 @@ spec:
         # More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
         resources:
           limits:
-            cpu: 500m
-            memory: 768Mi
+            cpu: 4000m
```
rooftopcellist, line 100: @lahavyuv86 One small change, could we make the cpu 2000m?

PR #208, opened 2024-06-06
Remove x-kubernetes-preserve-unknown-fields: true

rooftopcellist:
Remove x-kubernetes-preserve-unknown-fields: true so that `oc explain eda.spec` works.

**Without this change:**

```
$ oc explain eda.spec
KIND:     EDA
VERSION:  eda.ansible.com/v1alpha1

DESCRIPTION:
     Spec defines the desired state of EDA

```

**With this change:**

```
$ oc explain eda.spec
KIND:     EDA
VERSION:  eda.ansible.com/v1alpha1

RESOURCE: spec <Object>

DESCRIPTION:
     Spec defines the desired state of EDA

FIELDS:
   activation_worker	<Object>
     Defines desired state of eda-activation-worker resources

   additional_labels	<[]string>
     Additional labels defined on the resource, which should be propagated to
     child resources

   admin_password_secret	<string>
     Secret where the admin password can be found. If not specified, one will be
     generated.

   admin_user	<string>
     Username to use for the admin account

   api	<Object>
     Defines desired state of eda-api resources

   automation_server_ssl_verify	<string>
     SSL Verify (yes, no) for the AWX automation host

   automation_server_url	<string> -required-
     URL of the AWX automation host to run ansible jobs against

   bundle_cacert_secret	<string>
     Secret where the trusted Certificate Authority Bundle is stored

   database	<Object>
     The EDA PostgreSQL database StatefulSet

   db_fields_encryption_secret	<string>
     Secret where the DB fields encryption key can be found. If not specified,
     one will be generated.

   default_worker	<Object>
     Defines desired state of eda-default-worker resources

   extra_settings	<[]Object>
     Environment variables to configure the application-level settings

   force_drop_db	<boolean>
     Force drop the database before restoring. USE WITH CAUTION!

   hostname	<string>
     The hostname of the instance

   image	<string>
     Registry path to API container image to use

   image_pull_policy	<string>
     The image pull policy

   image_pull_secrets	<[]string>
     Image pull secrets for app and database containers

   image_version	<string>
     API container image tag or version to use

   image_web	<string>
     Registry path to UI container image to use

   image_web_version	<string>
     UI container image tag or version to use

   ingress_annotations	<string>
     Annotations to add to the Ingress Controller

   ingress_api_version	<string>
     The Ingress API version to use

   ingress_class_name	<string>
     The name of ingress class to use instead of the cluster default.

   ingress_path	<string>
     The ingress path used to reach the deployed service

   ingress_path_type	<string>
[...]

PR #270, opened 2025-01-21
Remove x-kubernetes-preserve-unknown-fields field from schema so explain params work

rooftopcellist:
Remove x-kubernetes-preserve-unknown-fields field from schema so that users can run:

```
oc explain EDA.spec
```

When I remove this line under EDA.spec in the CRD, the oc explain works:

```
x-kubernetes-preserve-unknown-fields: true
```

When this line is removed, or set to false (default), then parameters that are not explicitly defined on the CRD schema will not be accepted and the k8s validator will automatically (and silently) omit any parameters that it does not know about when the CRD is modified / created.

##### ADDITIONAL INFORMATION

In the case of EDA, I think that should be fine because all accepted params are defined in the CRD schema already.

The `spec.extra_settings` param will still set that preserve-unknown-fields config to true for just extra_settings, so that's OK.

### DIFF
--- a/config/crd/bases/eda.ansible.com_edabackups.yaml
+++ b/config/crd/bases/eda.ansible.com_edabackups.yaml
@@ -38,7 +38,6 @@ spec:
             type: object
           spec:
             type: object
-            x-kubernetes-preserve-unknown-fields: true
             required:
             - deployment_name
             properties:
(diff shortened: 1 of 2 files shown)


rooftopcellist:
Note, this was fixed for the EDA CR previously in:
* https://github.com/ansible/eda-server-operator/pull/208

The current PR is to fix it in backup and restore objects.

dsavineau (approved):

Issue #265, opened 2025-01-08
Event streams can't be created when public_base_url is not defined


Alex-Izquierdo:
It seems that `public_base_url` has no default value and the documentation doesn't not say that it must be defined, so if the application is installed without defining that variable event streams can not be created, the application returns an error like: `EventStream of type Token Event Stream cannot be used because EVENT_STREAM_BASE_URL is not configured. Please check with your site administrator.`

The documentation should indicate that the variable must be defined in order to work with event streams or it should have a default value. 

This issue has been reported in the forum. [Ref](https://forum.ansible.com/t/some-help-for-newbie-event-stream-base-url-is-not-configured-in-eda/39501/1).

Alex-Izquierdo:
For now, the goal is [to clarify in the documentation](https://github.com/ansible/eda-server-operator/pull/274) how to configure the CRD to get working event streams in a standalone deployment.  Ideally we should be able to handle a default value of `public_base_url` automatically based on the ingress. But a final solution needs to be discussed. The problem is that we can not set arbitrarily `public_base_url` since this variable is used also to determine that an external UI provider is set and it's existence would disable the standalone EDA UI, so the only way right now is to override manually the application setting `EDA_EVENT_STREAM_BASE_URL`

We need to implement another way to disable/enable the standalone UI and set `public_base_url` for standalone deployments.

PR #274, opened 2025-03-05
docs: add how to set event stream base url AAP-38345

Alex-Izquierdo:
By default, `EDA_EVENT_STREAM_BASE_URL` is not set, preventing event streams working in standalone deployments. This PR updates the documentation until a final solution to handle `public_base_url` is implemented. 

Ref: https://github.com/ansible/eda-server-operator/issues/265

Internal jira: https://issues.redhat.com/browse/AAP-38345

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,14 @@ spec:
 Optionally, it is also recommended to set `event_stream.prefix` with a non guessable value (for example an UUID) to avoid unwanted access to the event streams or DDOS attacks. 
 
 ```yaml
+apiVersion: eda.ansible.com/v1alpha1
+kind: EDA
+metadata:
+  name: eda
+spec:
+  event_stream:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -218,6 +218,25 @@ spec:
...
+kind: EDA
+metadata:
+  name: eda
+spec:
+  extra_settings:
+    - setting: EDA_EVENT_STREAM_BASE_URL
+      value: "https://mypublicdomain.com/eda-event-streams"
+```
+
+Optionally, it is also recommended to set `event_stream.prefix` with a non guessable value (for example an UUID) to avoid unwanted access to the event streams or DDOS attacks.
```
ttuffin, line 236: Where do you set `event_stream.prefix`? Should we elaborate on how to do that?

Alex-Izquierdo: updated.

ttuffin (approved):

zkayyali812 (approved):

Issue #276, opened 2025-03-14
Getting issue while installing eda-server-operator instance with 1.0.0 and main


prtkdaved:
│   File "/app/venv/lib64/python3.11/site-packages/django/core/management/__init__.py", line 262, in fetch_command                                                             │
│     settings.INSTALLED_APPS                                                                                                                                                  │
│   File "/app/venv/lib64/python3.11/site-packages/django/conf/__init__.py", line 102, in __getattr__                                                                          │
│     self._setup(name)                                                                                                                                                        │
│   File "/app/venv/lib64/python3.11/site-packages/django/conf/__init__.py", line 89, in _setup                                                                                │
│     self._wrapped = Settings(settings_module)                                                                                                                                │
│                     ^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                │
│   File "/app/venv/lib64/python3.11/site-packages/django/conf/__init__.py", line 217, in __init__                                                                             │
│     mod = importlib.import_module(self.SETTINGS_MODULE)                                                                                                                      │
│           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                      │
│   File "/usr/lib64/python3.11/importlib/__init__.py", line 126, in import_module                                                                                             │
│     return _bootstrap._gcd_import(name[level:], package, level)                                                                                                              │
│            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                              │
│   File "<frozen importlib._bootstrap>", line 1204, in _gcd_import                                                                                                            │
│   File "<frozen importlib._bootstrap>", line 1176, in _find_and_load                                                                                                         │
│   File "<frozen importlib._bootstrap>", line 1126, in _find_and_load_unlocked                                                                                                │
│   File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed                                                                                               │
│   File "<frozen importlib._bootstrap>", line 1204, in _gcd_import                                                                                                            │
│   File "<frozen importlib._bootstrap>", line 1176, in _find_and_load                                                                                                         │
│   File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked                                                                                                │
│   File "<frozen importlib._bootstrap>", line 690, in _load_unlocked                                                                                                          │
│   File "<frozen importlib._bootstrap_external>", line 940, in exec_module                                                                                                    │
[...]

PR #277, opened 2025-03-14
Update kubernetes.core to 2.4.2 to fix k8s_cp module usage against OCP with virt

rooftopcellist:
Resolves a backup bug hit when copying the CR object yaml to the backup PVC.

Fixes the k8s_cp task usage, which fails on newer Openshift clusters that have Openshift Virtualization installed.

Related k8s core issue:

>   - Resolve Collections util resource discovery fails when complex subresources
>     present (https://github.com/ansible-collections/kubernetes.core/pull/676).


dsavineau (approved):

rooftopcellist:
CI for this PR is failing because of a a CI issue caused by the recent settings changes in the eda-server application. 
* https://github.com/ansible/eda-server-operator/issues/276

These changes were tested manually against an older version of the eda-server app image and worked.

PR #286, opened 2025-05-07
Add the ability to migrate from ClusterA to ClusterB

rooftopcellist:
This new logic (roles/eda/tasks/migrate_data.yml) adds the ability to migrate the database of an EDA deployment from one cluster to another. 

### EDA Migration Steps: Cluster A to Cluster B
1. Prepare Source Database Information:
  * Create a secret with your old DB fields encryption key
  * Create a secret with your old PostgreSQL connection details
2. Deploy EDA on New Cluster:
  * Create EDA CR referencing both secrets in the spec
3. Migration Process:
  * The operator automatically handles data migration
  * Streams data from old to new PostgreSQL instance
  * Runs database migrations as needed
4. Post-Migration:
  * Operator automatically removes old database reference
  * New EDA instance becomes operational with migrated data

Note: This only migrates the database and any configuration that you have on your EDA CR.  It is necessary to recreate any k8s secrets used by the EDA CR in the new namespace on clusterB.

This same procedure could be used to migrate from one namespace to another on the same cluster.

### DIFF
--- a/config/crd/bases/eda.ansible.com_edas.yaml
+++ b/config/crd/bases/eda.ansible.com_edas.yaml
@@ -2829,6 +2829,18 @@ spec:
               idle_deployment:
                 description: Scale down deployments to put EDA into an idle state
                 type: boolean
+              old_postgres_configuration_secret:
+                description: Secret where the old database configuration can be found for data migration
+                type: string
+                maxLength: 255
+                pattern: '^[a-zA-Z0-9][-a-zA-Z0-9]{0,253}[a-zA-Z0-9]$'
+              pg_dump_suffix:
+                description: Additional parameters for the pg_dump command during a migration
+                type: string
+              postgres_label_selector:
+                description: Label selector used to identify postgres pod for data migration
+                type: string
+
           status:
             description: Status defines the observed state of EDA
             properties:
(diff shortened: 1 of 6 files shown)


Alex-Izquierdo (approved):
LGTM but the name of "pg_dump_suffix" variable is a standard in our operators? Because I feel it is some misleading when it is used to append arguments for pg_dump. I would suggest something more self descriptive like "pg_dump_extra_params" or similar.

rooftopcellist:
@Alex-Izquierdo thanks for the review!  Yes that `pg_dump_suffix` pattern is already in the awx-operator.  I think we should use the name name for consistency.
* https://github.com/ansible/awx-operator/blob/54072d6a46dcb0b4ed75a6af810566493c1e7230/roles/installer/tasks/migrate_data.yml#L47

rooftopcellist:
# Testing

1. Build and deploy operator with up.sh script in the `eda` namespace with name `eda`
2. Create a new namespace called eda-2
3. Edit the password in the old-eda-postgres-configuration secret
4. Create the old-db-fields-encryption and old-postgres-configuration secrets in that namespace, as described in the `docs/migration/migration.md` guide.

```
oc apply -f secrets/
```


Note: In the old postgresql secret, if going from one namespace to another, you can set the host like this to make it routable within the cluster without needing to make a Route:

```
host: eda-2-postgres-15.eda-2.svc.cluster.local
```

5. Deploy the eda-server-operator to the new eda-2 namespace (comment out up.sh block that creates the CR, then run it it the root of the clone).
6. Create the new eda-migrated CR

```
oc create -f eda-migrated-openshift-cr.yml
```

Using these steps, I was able to migrate from the `eda` namespace to the `eda-2` namespace.  This same approach would work for migrating from cluster to cluster, it is just necessary to expose the database hostname.

PR #287, opened 2025-05-13
Add status.conditions.ansibleResult to the CRD and CSV to fix unknown status warning

rooftopcellist:
Without this change, the status condition is unknown and this warning is found in the operator logs:

```
  {"level":"info","ts":1747091990.8347256,"logger":"KubeAPIWarningLogger","msg":"unknown field status.conditions[1].ansibleResult}
```

We should do this in all of the operators.

### DIFF
--- a/config/crd/bases/eda.ansible.com_edabackups.yaml
+++ b/config/crd/bases/eda.ansible.com_edabackups.yaml
@@ -113,6 +113,9 @@ spec:
                       type: string
                     type:
                       type: string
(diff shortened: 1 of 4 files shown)


ttuffin (approved):

dsavineau (approved):

rooftopcellist:
FYI, I updated the PR to add these same changes to the backup and restore CRD's and those sections of the CSV so that this is fixed everywhere.

PR #292, opened 2025-07-02
Make the down.sh script exclude cleaning up CRDs by default

rooftopcellist:
# Summary

Without this change, by default, the down.sh will delete all EDA related CRD's on the cluster.  If there are other EDA operators on the cluster or the AAP Operator is in another namespace on this cluster, this could clobber the CRD for all.  To avoid that, I am changing the default behavior so that it keeps the CRD's.  Optionally, you can pass the `--all` flag to delete them as well.

## Details

Added --all flag to  down.sh
This PR adds an optional --all flag to the  down.sh script that controls whether CRDs are removed during cleanup.

Usage:
 * ./down.sh - Default behavior, uses make undeploy-keep-crd to clean up resources while preserving CRDs
* ./down.sh --all - Enhanced cleanup, uses make undeploy to remove all resources including CRDs
This change helps developers who need a complete cleanup of their environment, while maintaining backward compatibility with the existing workflow.

```
./down.sh --all
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,6 +122,11 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 	cd config/default && $(KUSTOMIZE) edit set namespace ${NAMESPACE}
 	$(KUSTOMIZE) build config/default | kubectl delete -f -
 
+.PHONY: undeploy-keep-crd
+undeploy-keep-crd: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Prevents down.sh from stomping on other CRD's in the same cluster.
+	cd config/default-keep-crd && $(KUSTOMIZE) edit set namespace ${NAMESPACE}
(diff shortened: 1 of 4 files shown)


Alex-Izquierdo (approved):
LGTM, should we document it in https://github.com/ansible/eda-server-operator/blob/main/docs/development.md#clean-up ?

rooftopcellist:
Closing this out. `down.sh` was removed entirely in c80f3fc ("Add standardized Makefile includes and developer documentation"), which replaced the old script-based dev workflow with `make up` / `make down` (`makefiles/common.mk` + `makefiles/operator.mk`, documented in `docs/development.md`).

The concern this PR addressed is already handled by the new workflow, and more thoroughly than the proposed `--all` flag: `make down`'s teardown chain never touches CRDs at all. CRD removal is now a fully separate, explicit `make uninstall` target, so there's no risk of accidentally clobbering shared CRDs on teardown.

No further action needed here.

PR #295, opened 2025-07-07
Update document on how to allow manage resources by EDA

bzwei:
needed by https://github.com/ansible/eda-server/pull/1359

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,9 @@ Table of Contents
   - [Upgrades](#upgrades)
   - [Advanced Configuration](#advanced-configuration)
     - [Admin user account configuration](#admin-user-account-configuration)
+    - [Allow Local Resource Management](#allow-local-resource-management)
     - [Database Fields Encryption Configuration](#database-fields-encryption-configuration)
+    - [Enable Event Streams](#enable-event-streams)
     - [Additional Advanced Configuration](#additional-advanced-configuration)
   - [Maintainers Docs](#maintainers-docs)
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -183,6 +185,21 @@ stringData:
 ```
 
 
+### Allow Local Resource Management
+
+Resources such as users, teams, or organizations are recommended to be managed by the gateway. Flag `EDA_ALLOW_LOCAL_RESOURCE_MANAGEMENT` can be toggled to instruct whether EDA can create/modify/delete these resources. The following example allows EDA to manage such resources. It is useful when EDA is deployed alone.
```
Alex-Izquierdo, line 190: I think the right term for upstream is "platform provider".
```suggestion
Resources such as users, teams, or organizations are recommended to be managed by the *Platform Provider*.  Flag `EDA_ALLOW_LOCAL_RESOURCE_MANAGEMENT` can be toggled to instruct whether EDA can create/modify/delete these resources. The following example allows EDA to manage such resources. It is necessary when EDA is deployed alone.
```

dev/eda-cr/eda-k8s-ing.yml
```diff
@@ -27,6 +27,8 @@ spec:
   extra_settings:
     - setting: EDA_MAX_RUNNING_ACTIVATIONS
       value: '11'
+    - setting: EDA_ALLOW_LOCAL_RESOURCE_MANAGEMENT
```
Alex-Izquierdo, line 30: There are two more files in `dev/` they should not have this setting too? in the end all are standalone deployments.

Alex-Izquierdo (approved):

PR #307, opened 2025-11-14
Adding affinity parameter in deployments

aldato:
# Summary

This PR adds the capability of managing the affinity of the different component deployments being used by EDA. 

## Changes

- Added the necessary changes in the EDA CDR
- Added the defaults for affinity
- Added the required conditionals in the Jinja2 templates

## Tests

I tested it using simple podAntiAffinity rules within the EDA CR under spec.event_stream with the following and it worked just fine (against the task pod of controller):

```
    task_affinity:
      podAntiAffinity:
        requiredDuringSchedulingIgnoredDuringExecution:
        - labelSelector:
            matchExpressions:
            - key: app.kubernetes.io/name
              operator: In
              values:
              - automation-controller-task
          topologyKey: kubernetes.io/hostname
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Added support for optional scheduling affinity settings across EDA deployments, including API, workers, event stream, and UI.
  * Corrected template rendering so scheduling rules are applied independently and consistently when set.
  * Ensured default configuration now includes an empty affinity section for all relevant deployment blocks.

### DIFF
--- a/roles/eda/defaults/main.yml
+++ b/roles/eda/defaults/main.yml
@@ -38,6 +38,7 @@ _api:
       memory: 350Mi
   node_selector: {}
   tolerations: []
+  affinity: {}
 
 # Note: "default-worker: {}" is intentionally excluded here so we know if the user set it
 _default_worker:
@@ -48,6 +49,7 @@ _default_worker:
       memory: 130Mi
   node_selector: {}
   tolerations: []
+  affinity: {}
 
 # Note: "activation-worker: {}" is intentionally excluded here so we know if the user set it
 _activation_worker:
@@ -58,6 +60,7 @@ _activation_worker:
       memory: 150Mi
   node_selector: {}
   tolerations: []
+  affinity: {}
 
 # Note: Deprecated "worker: {}" is intentionally excluded here so we know if the user set it
 _worker: {}
@@ -73,6 +76,7 @@ _event_stream:
       memory: 350Mi
   node_selector: {}
   tolerations: []
+  affinity: {}
 
 ui: {}
 _ui:
@@ -83,6 +87,7 @@ _ui:
       memory: 64Mi
   node_selector: {}
   tolerations: []
+  affinity: {}
 
 # Labels defined on the resource, which should be propagated to child resources
 additional_labels: []
--- a/config/crd/bases/eda.ansible.com_edas.yaml
+++ b/config/crd/bases/eda.ansible.com_edas.yaml
@@ -440,6 +440,372 @@ spec:
                       - whenUnsatisfiable
                       type: object
                     type: array
+                  affinity:
+                    description: If specified, the pod's scheduling constraints
+                    properties:
+                      nodeAffinity:
+                        properties:
+                          preferredDuringSchedulingIgnoredDuringExecution:
+                            items:
+                              properties:
+                                preference:
+                                  properties:
+                                    matchExpressions:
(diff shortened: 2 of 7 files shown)


roles/eda/templates/eda-scheduler.deployment.yaml.j2
```diff
@@ -62,6 +62,10 @@ spec:
 {% if combined_scheduler.topology_spread_constraints is defined %} %}
       topologySpreadConstraints:
         {{ combined_scheduler.topology_spread_constraints | indent(width=8) }}
+{% endif %}
+{% if combined_scheduler.affinity is defined and combined_scheduler.affinity | length > 0 %}
```
rooftopcellist: **[Major — Functionality]** This template references `combined_scheduler.affinity`, and `_scheduler` in `defaults/main.yml` defines `affinity: {}`, but the **CRD is missing the `affinity` field in the `scheduler` section**.

The CRD adds `affinity` to 6 component sections (api, event_stream, ui, default_worker, activation_worker, worker), but the `scheduler` section (CRD line ~4297) ends after `topology_spread_constraints` with no `affinity` property.

This means users who set `spec.scheduler.affinity` in their CR will have the field **silently stripped** by CRD validation — scheduler pods deploy without the requested affinity, with no error or warning.

**Fix:** Add the identical affinity schema block (366 lines, matching the pattern in the other component sections) to the `scheduler` section in `config/crd/bases/eda.ansible.com_edas.yaml`, after `topology_spread_constraints`.

roles/eda/defaults/main.yml
```diff
@@ -38,6 +38,7 @@ _api:
       memory: 350Mi
   node_selector: {}
   tolerations: []
+  affinity: {}
```
rooftopcellist, line 41: **[Minor — Quality]** The sample CR (`config/samples/eda_v1alpha1_eda.yaml`) was not updated to include an affinity example. Since this is a new user-facing scheduling feature, a commented-out example would help discoverability. The `podAntiAffinity` test config from the PR description would work well as a sample:

```yaml
# affinity:
#   podAntiAffinity:
#     requiredDuringSchedulingIgnoredDuringExecution:
#     - labelSelector:
#         matchExpressions:
#         - key: app.kubernetes.io/name
#           operator: In
#           values:
#           - automation-controller-task
#       topologyKey: kubernetes.io/hostname
```

rooftopcellist (commented):
Thanks for adding affinity support across the EDA components — clean implementation that follows the right patterns. Two items to address before this is ready.

roles/eda/templates/eda-scheduler.deployment.yaml.j2
aldato: Hey @rooftopcellist, thanks for checking this! At some point between the time I initially put the PR and today, I believe the scheduler component got entirely removed in main: https://github.com/ansible/eda-server-operator/blob/main/roles/eda/tasks/deploy_eda.yml#L95-L101 . I removed the scheduler traces in the PR.

rooftopcellist: Great! Thanks for the updates. It is looking good now.

rooftopcellist (approved):

rcarrillocruz (approved):

rooftopcellist (approved):

PR #331, opened 2026-03-24
fix: gate RESOURCE_SERVER and local resource management on gateway presence

kaiokmo:
EDA server defaults `RESOURCE_SERVER__URL` to `"https://localhost"`, which causes `apply_resource_server_auth` to unconditionally restrict authentication to JWT-only, breaking session-based login for standalone deployments. This was introduced by the gateway-only auth enforcing, whenever `RESOURCE_SERVER__URL` is truthy. Development mode is unaffected because development_defaults.py sets it to None [3].

We now use `public_base_url` (the same mechanism other operators use) to detect gateway presence and explicitly configure standalone deployments in the ConfigMap. It should unblock access to the UI when deploying the EDA operator in standalone mode.

## New Features
* Added automatic local resource management when no public base URL is configured.
* Resource server settings are now applied conditionally based on the configured public base URL.

## Documentation
* Updated standalone deployment guidance to reflect the automatic configuration behavior.

## Configuration
* Simplified deployment examples by removing redundant local resource management settings.

### DIFF
--- a/roles/eda/templates/eda.configmap.yaml.j2
+++ b/roles/eda/templates/eda.configmap.yaml.j2
@@ -35,9 +35,14 @@ data:
   EDA_STATIC_URL: /api/eda/static/
 
   # Resource Server configuration
-{% if not (public_base_url | default('') | length > 0) %}
+{% set _user_settings = (extra_settings | default([])) | map(attribute='setting') | map('upper') | list %}
+{% if public_base_url | default('') | length == 0 %}
+{% if 'EDA_RESOURCE_SERVER__URL' not in _user_settings %}
   EDA_RESOURCE_SERVER__URL: ""
+{% endif %}
+{% if 'EDA_ALLOW_LOCAL_RESOURCE_MANAGEMENT' not in _user_settings %}
   EDA_ALLOW_LOCAL_RESOURCE_MANAGEMENT: "True"
+{% endif %}
 {% endif %}
 
   # Custom user variables
--- a/roles/eda/templates/eda.configmap.yaml.j2
+++ b/roles/eda/templates/eda.configmap.yaml.j2
@@ -34,6 +34,12 @@ data:
 
   EDA_STATIC_URL: /api/eda/static/
 
+  # Resource Server configuration
+{% if not (public_base_url | default('') | length > 0) %}
+  EDA_RESOURCE_SERVER__URL: ""
+  EDA_ALLOW_LOCAL_RESOURCE_MANAGEMENT: "True"
+{% endif %}
+
   # Custom user variables
 {% for item in extra_settings | default([]) %}
   {{ item.setting | upper }}: "{{ item.value }}"
--- a/README.md
+++ b/README.md
@@ -187,7 +187,9 @@ stringData:
 
 ### Allow Local Resource Management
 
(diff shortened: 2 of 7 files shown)


jon-nfc:
Is this going to be merged anytime soon? This PR is a blocker to ansible/eda-server#1504

jon-nfc:
bump,

can you provide an update to when this pr will be addressed. This pr is a blocker to https://github.com/ansible/eda-server/issues/1504 as already stated.

Additionally now due to lack of action, this PR blocks https://github.com/ansible/ansible-ui/pull/3322 as no newer eda-server can be deployed to check/confirm this PR.

jon-nfc:
bump

roles/eda/templates/eda.configmap.yaml.j2
```diff
@@ -34,6 +34,12 @@ data:
 
   EDA_STATIC_URL: /api/eda/static/
 
+  # Resource Server configuration
+{% if not (public_base_url | default('') | length > 0) %}
+  EDA_RESOURCE_SERVER__URL: ""
+  EDA_ALLOW_LOCAL_RESOURCE_MANAGEMENT: "True"
+{% endif %}
```
rooftopcellist, line 46: The new block can end up emitting `EDA_ALLOW_LOCAL_RESOURCE_MANAGEMENT` twice in the rendered ConfigMap: every bundled standalone sample (`dev/eda-cr/eda-k8s-ing.yml`, `eda-openshift-cr.yml`, `lightweight-eda.yml`, `eda-k8s-nodeport-cr.yml`, `eda-resource-quota-cr.yml`) already sets this key via `extra_settings`, and none of them set `public_base_url`, so both this block and the `extra_settings` loop below fire for all of them. `kubernetes.core`'s `k8s` module parses the rendered manifest with `yaml.safe_load_all()`, which silently keeps the last occurrence of a duplicate key. Harmless today only because `extra_settings` renders after this block and happens to win with the same value - it stops being harmless if the block order ever changes or a CR sets a differing value.

Suggest guarding the new keys against ones already declared in `extra_settings`. Also fixes the double-negative condition while we're in here (equivalent to the positive guard already used in `roles/eda/tasks/deploy_eda.yml:20`):

```suggestion
  # Resource Server configuration
{% set _user_settings = (extra_settings | default([])) | map(attribute='setting') | map('upper') | list %}
{% if public_base_url | default('') | length == 0 %}
{% if 'EDA_RESOURCE_SERVER__URL' not in _user_settings %}
  EDA_RESOURCE_SERVER__URL: ""
{% endif %}
{% if 'EDA_ALLOW_LOCAL_RESOURCE_MANAGEMENT' not in _user_settings %}
  EDA_ALLOW_LOCAL_RESOURCE_MANAGEMENT: "True"
{% endif %}
{% endif %}
```

rooftopcellist:
Now that standalone deployments (no `public_base_url`) get `EDA_ALLOW_LOCAL_RESOURCE_MANAGEMENT` defaulted automatically, worth a follow-up to update `README.md`'s "Allow Local Resource Management" section and drop the now-redundant `extra_settings` entries from the bundled samples (`dev/eda-cr/eda-k8s-ing.yml`, `eda-openshift-cr.yml`, `lightweight-eda.yml`, `eda-k8s-nodeport-cr.yml`, `eda-resource-quota-cr.yml`) - that's also what's causing the duplicate key noted in the inline comment above.

Non-blocking either way - the core fix (using `public_base_url` to detect gateway presence, same mechanism used elsewhere in this role) looks right to me.

rooftopcellist (approved):

PR #345, opened 2026-05-04
feat: add configurable namespace for activation job pods

### LINKED ISSUE #344: feat: support configurable namespace for activation job pods
## Summary

When a rulebook activation fires, the EDA activation worker creates a Kubernetes Job in the same namespace where EDA is deployed. There is currently no way to direct those activation job pods into a separate namespace, which is a common requirement for security isolation, resource quota management, and NetworkPolicy boundaries.

## Current behaviour

The `kubernetes.Engine._set_namespace()` method in `eda-server` reads the namespace from the ServiceAccount token file (`/var/run/secrets/kubernetes.io/serviceaccount/namespace`). All Jobs, Pods, Secrets, and Services created by the activation worker land in the EDA namespace. The operator CRD's `activation_worker.node_selector` and `activation_worker.tolerations` only control the worker Deployment itself, not the spawned job pods.

## Proposed behaviour

A new optional CRD field (e.g. `activation_worker.activation_job_namespace`) that, when set:

### 1. Operator side (`eda-server-operator`)

- Injects the target namespace as an environment variable (e.g. `EDA_ACTIVATION_JOB_NAMESPACE`) into the activation worker Deployment via the EDA configmap.
- Creates a RoleBinding in the target namespace granting the EDA ServiceAccount permissions to manage Jobs, Pods, Secrets, and Services (the same verbs the existing activation worker role provides).
- Optionally creates the target Namespace if it does not exist.

### 2. Server side (`eda-server`)

- `_set_namespace()` reads the env var as an override, falling back to the SA token file.
- All `create_namespaced_*` and `list_namespaced_*` calls use the configured namespace.

## Motivation

- **Security isolation:** activation job pods run user-supplied rulebooks and decision environments. Separating them from the EDA control plane reduces blast radius.
- **Resource quotas:** a dedicated namespace allows independent ResourceQuota and LimitRange policies for activation workloads.
- **NetworkPolicy:** cross-namespace policies can restrict activation pods to only communicate with the EDA API (websocket callback) and required external services.

## Implementation considerations

- The EDA CRD is **namespace-scoped**, so cross-namespace RBAC must be managed explicitly by the operator.
- Activation job pods connect back to the EDA API via websocket. NetworkPolicies must allow this traffic path.
- The operator reconcile loop must keep the RoleBinding in sync (create on first reconcile, update if the SA name changes, clean up on CR deletion).
- Related upstream server PRs: [ansible/eda-server#1520](https://github.com/ansible/eda-server/pull/1520) (pod metadata), [ansible/eda-server#1545](https://github.com/ansible/eda-server/pull/1545) (tolerations), [ansible/eda-server#1534](https://github.com/ansible/eda-server/pull/1534) (resource limits).
- Related RFE: AAPRFE-2827.

amasolov:
## Summary

- Adds `activation_job_namespace` field to the `activation_worker` section of the EDA CRD, allowing activation job pods to run in a separate Kubernetes namespace for security isolation, resource quota management, and NetworkPolicy boundaries.
- The operator injects `EDA_ACTIVATION_JOB_NAMESPACE` into the activation worker ConfigMap and creates cross-namespace RBAC so the EDA ServiceAccount can manage Jobs, Pods, Secrets, and Services in the target namespace.

## What changed

- **CRD**: new optional string field `activation_worker.activation_job_namespace`.
- **Role defaults**: `activation_job_namespace` defaults to empty string (feature disabled).
- **ConfigMap template**: conditionally sets `EDA_ACTIVATION_JOB_NAMESPACE` env var.
- **New RBAC template**: `eda-activation-job-namespace-rbac.yaml.j2` creates a Role and RoleBinding in the target namespace.
- **Deploy task**: applies cross-namespace RBAC when `activation_job_namespace` is set; removes it when unset.
- **ClusterRole + ClusterRoleBinding**: grants the operator SA permission to manage Roles and RoleBindings in the target namespace.

## Why

Operators running EDA on shared clusters need activation job pods (which run user-supplied rulebooks and decision environments) isolated from the EDA control plane. A dedicated namespace enables independent ResourceQuota, LimitRange, and NetworkPolicy controls.

## Server-side companion

The `eda-server` side reads `EDA_ACTIVATION_JOB_NAMESPACE` as an override in `_set_namespace()`, falling back to the SA token file. A companion PR will be filed on `ansible/eda-server`.

## How to test

1. Deploy an EDA CR without `activation_job_namespace` (default behaviour, no change).
2. Set `activation_worker.activation_job_namespace: "eda-jobs"` on the CR.
3. Verify the operator creates Role + RoleBinding in the `eda-jobs` namespace.
4. Verify `EDA_ACTIVATION_JOB_NAMESPACE` appears in the activation worker ConfigMap.

## Breaking changes / dependencies

None. Field is optional; existing CRs are unaffected.

Closes #344.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Specify a custom Kubernetes namespace for activation job execution.
  * Automatically provision and clean up required cross-namespace permissions when the configured namespace changes.
  * Network access policies now appropriately restrict activation job traffic to the configured namespace.

* **Tests**
  * Added a CI scenario to validate the activation-job-namespace workflow.

### DIFF
--- a/roles/eda/tasks/deploy_eda.yml
+++ b/roles/eda/tasks/deploy_eda.yml
@@ -38,6 +38,44 @@
     wait: yes
   when: public_base_url is defined
 
+- name: Look up existing ConfigMap for previous activation_job_namespace
+  kubernetes.core.k8s_info:
+    api_version: v1
+    kind: ConfigMap
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    name: "{{ ansible_operator_meta.name }}-{{ deployment_type }}-env-properties"
+  register: _eda_env_cm
+
+- name: Record previous activation_job_namespace from ConfigMap
+  ansible.builtin.set_fact:
+    _previous_activation_job_namespace: >-
+      {{ (_eda_env_cm.resources | first).data.EDA_ACTIVATION_JOB_NAMESPACE | default('') }}
+  when:
+    - _eda_env_cm.resources | length > 0
+    - (_eda_env_cm.resources | first).data is defined
+
+- name: Apply cross-namespace RBAC for activation job pods
+  k8s:
+    apply: yes
+    definition: "{{ lookup('template', 'eda-activation-job-namespace-rbac.yaml.j2') }}"
+    wait: yes
+  when: combined_activation_worker.activation_job_namespace | default('') | length > 0
+
+- name: Remove cross-namespace RBAC when activation_job_namespace is unset
+  k8s:
+    state: absent
+    api_version: rbac.authorization.k8s.io/v1
+    kind: "{{ item.kind }}"
+    name: "{{ ansible_operator_meta.name }}-activation-job-manager"
+    namespace: "{{ _previous_activation_job_namespace }}"
+  loop:
+    - { kind: RoleBinding }
+    - { kind: Role }
+  when:
+    - combined_activation_worker.activation_job_namespace | default('') | length == 0
+    - _previous_activation_job_namespace | default('') | length > 0
+  ignore_errors: yes
+
 - name: Apply Backend deployment resources
   k8s:
     apply: yes
--- a/roles/eda/tasks/deploy_eda.yml
+++ b/roles/eda/tasks/deploy_eda.yml
@@ -19,6 +19,22 @@
     event_stream_mtls_base_url: "{{ public_base_url.rstrip('/') }}/{{ event_stream_mtls_prefix_path }}"
   when: public_base_url | default('') | length > 0
 
+- name: Look up existing ConfigMap for previous activation_job_namespace
+  kubernetes.core.k8s_info:
+    api_version: v1
+    kind: ConfigMap
+    namespace: "{{ ansible_operator_meta.namespace }}"
+    name: "{{ ansible_operator_meta.name }}-{{ deployment_type }}-env-properties"
+  register: _eda_env_cm
+
+- name: Record previous activation_job_namespace from ConfigMap
+  ansible.builtin.set_fact:
+    _previous_activation_job_namespace: >-
+      {{ (_eda_env_cm.resources | first).data.EDA_ACTIVATION_JOB_NAMESPACE | default('') }}
+  when:
+    - _eda_env_cm.resources | length > 0
+    - (_eda_env_cm.resources | first).data is defined
+
 - name: Apply ConfigMap resources
   k8s:
     apply: yes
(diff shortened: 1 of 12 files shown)


ttuffin:
@amasolov I think it would be worthwhile adding a new CR (e.g `.ci/eda_v1alpha1_eda.activation_job_namespace.ci.yaml`) to the CI to test this new parameter.

```
activation_worker:
  activation_job_namespace: "eda-jobs"
```

And then add it to the matrix in pr.yml:
```
- SCENARIO: activation_job_namespace
```

We'll also need to add a step to ensure the namespace is create before applying the CR:
```
- name: Create activation job namespace
  run: kubectl create namespace eda-jobs
  if: ${{ matrix.SCENARIO == 'activation_job_namespace' }}
```

amasolov:
> @amasolov I think it would be worthwhile adding a new CR 

Agreed and done!

mkanoor (approved):

roles/eda/tasks/deploy_eda.yml
```diff
@@ -38,6 +54,28 @@
...
+    kind: "{{ item.kind }}"
+    name: "{{ ansible_operator_meta.name }}-activation-job-manager"
+    namespace: "{{ _previous_activation_job_namespace }}"
+  loop:
+    - { kind: RoleBinding }
+    - { kind: Role }
+  when:
+    - _previous_activation_job_namespace | default('') | length > 0
+    - combined_activation_worker.activation_job_namespace | default('') != _previous_activation_job_namespace
+  ignore_errors: yes
```
kaiokmo: @amasolov I don't understand the reason for setting ignore_errors to true on this task. won't the module's `absent` idempotency be enough, along with the when guards?

amasolov: @kaiokmo You're right, state: absent already handles the "not found" case gracefully, and the when guards prevent other runs. Removed ignore_errors so legitimate failures (permission issues, API errors) surface properly.

roles/eda/tasks/deploy_eda.yml
```diff
@@ -38,6 +54,27 @@
     wait: yes
   when: public_base_url is defined
 
+- name: Apply cross-namespace RBAC for activation job pods
+  k8s:
+    apply: yes
+    definition: "{{ lookup('template', 'eda-activation-job-namespace-rbac.yaml.j2') }}"
+    wait: yes
+  when: combined_activation_worker.activation_job_namespace | default('') | length > 0
```
kaiokmo: #362 just merged NetworkPolicies with default-deny. The `eda-api` and `postgres` NP templates allow activation job pods via bare `podSelector: {app: eda}`, which won't match pods in a different namespace.

You'll need to add conditional `namespaceSelector` to both NP templates when `activation_job_namespace` is set, e.g.:

```yaml
{% if combined_activation_worker.activation_job_namespace | default('') | length > 0 %}
        - namespaceSelector:
            matchLabels:
              kubernetes.io/metadata.name: '{{ combined_activation_worker.activation_job_namespace }}'
          podSelector:
            matchLabels:
              app: eda
{% else %}
        - podSelector:
            matchLabels:
              app: eda
{% endif %}
```

Without this, rulebook execution (websocket), token refresh (API), and event stream activations (pg_notify) will fail when jobs run in a separate namespace.

kaiokmo (commented):
@amasolov this one needs a rebase against `main`. See inline comment about a NetworkPolicy gap from #362. Would be great if you could take care of it.

roles/eda/tasks/deploy_eda.yml
amasolov: @kaiokmo Done. Rebased against `main`, added conditional `namespaceSelector` to both the `eda-api` and `postgres` NetworkPolicy templates for cross-namespace activation job pods, and reordered the deploy tasks so the ConfigMap update runs after the RBAC migration. All checks are green.

kaiokmo (approved):
Awesome work, @amasolov. I'm approving these changes. You'll only have to deal with another rebase, and I think that's it.

amasolov:
@kaiokmo everything seems to be green after the rebase!

PR #350, opened 2026-05-27
remove: Remove deprecated reference to manager_auth_proxy_patch.yaml

Francisco-xiq:
## Summary

Removes a stale reference to `../default/manager_auth_proxy_patch.yaml` in `config/testing/kustomization.yaml`.

This file was deleted in #323 but the reference was left behind, causing `kustomize build` to fail for anyone using the `config/testing` directory.

Closes #262

## Changes

- `config/testing/kustomization.yaml`: removed path entry pointing to deleted file `../default/manager_auth_proxy_patch.yaml`

## Note

`make lint` was already failing before this change on `roles/postgres/` and `roles/restore/`. 

## Summary by CodeRabbit

* **Chores**
  * Updated testing environment configuration to switch manager image reference and enable more verbose debug logging during tests.
  * Removed an obsolete proxy patch to streamline test setup and improve configuration consistency.

PR #351, opened 2026-05-28
ci: Add workflow to validate all kustomize builds

Francisco-xiq:
## Summary

Adds a new GitHub Actions workflow to validate all `kustomization.yaml` files in the repository on every pull request.

This was suggested in #350, where a stale reference to a deleted file (`../default/manager_auth_proxy_patch.yaml`) went undetected until manually tested. A kustomize validation step was suggested to solve this problem

## What it does

- Finds all `kustomization.yaml` files in the repository (excluding `vendor/`)
- Runs `kustomize build` on each one to ensure all referenced files exist and the YAML is valid
- Reports all failures at once instead of stopping at the first error

## Testing

Validated locally and in CI on a test [PR](https://github.com/Francisco-xiq/eda-server-operator/pull/3) that intentionally reintroduced the broken reference, the workflow correctly failed with a clear error message pointing to the problematic file. 

## Summary by CodeRabbit

* **Chores**
  * Added an automated pull-request CI check that validates kustomize configurations across the repository (skips vendored paths). It runs per-directory build validations on pull requests, reports per-directory results, annotates failing manifests inline, and fails the PR check if any validation fails.

### DIFF
--- a/.github/workflows/validate-kustomize.yml
+++ b/.github/workflows/validate-kustomize.yml
@@ -1,0 +1,37 @@
+---
+name: Validate Kustomize
+
+on:
+  pull_request:
+
+jobs:
+  validate-kustomize:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      - name: Install kustomize
+        run: |
+          set -euxo pipefail
+          make kustomize
+          echo "$(pwd)/bin" >> $GITHUB_PATH
+
+      - name: Validate all kustomization.yaml files
+        run: |
+          set -euxo pipefail
+          failed=0
+          while IFS= read -r f; do
+            dir=$(dirname "$f")
+            echo "--- Validating $dir ---"
+            if ! kustomize build "$dir" > /dev/null; then
+              echo "::error file=$f::kustomize build failed for $dir"
+              failed=1
+            else
+              echo "✓ $dir OK"
+            fi
+          done < <(find . -name "kustomization.yaml" -not -path "*/vendor/*")
+          exit $failed


.github/workflows/validate-kustomize.yml
```diff
@@ -0,0 +1,33 @@
...
+on:
+  pull_request:
+
+jobs:
+  validate-kustomize:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
```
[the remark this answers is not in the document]
Francisco-xiq: Working on that

.github/workflows/validate-kustomize.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      - name: Install kustomize
+        run: |
+          curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
+          sudo mv kustomize /usr/local/bin/
```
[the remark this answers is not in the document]
Francisco-xiq: Working on that

Francisco-xiq:
Updated the workflow to use `make kustomize` so the CI stays in sync with whatever version the Makefile pins no duplication needed.

For the actions/checkout SHA pin, I've noticed that all other workflows here use @v4, so I kept it consistent. Happy to change if the project wants to adopt another solution.

.github/workflows/validate-kustomize.yml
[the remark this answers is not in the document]
Francisco-xiq: Updated the solution to use make kustomize instead of a manual curl install, so the CI stays in sync with the Makefile's pinned version automatically.
Regarding the actions/checkout SHA pin, the other workflows in this repo don't use it either, so I'll wait for a maintainer's before changing it.

.github/workflows/validate-kustomize.yml
```diff
@@ -0,0 +1,33 @@
+---
+name: Validate Kustomize
+
+on:
+  pull_request:
+
+jobs:
+  validate-kustomize:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
```
ptoscano: Make it a bit more loose, e.g. 10 minutes. This way it is more consistent with other workflows, and avoids timeouts due to slow github infra.

.github/workflows/validate-kustomize.yml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+
+      - name: Install kustomize
+        run: |
```
ptoscano, line 18: Add
```bash
set -euxo pipefail
```
as first instruction in the shell block: without it, if `make kustomize` fail it will not cause the failure of the step (and thus of the whole job).

.github/workflows/validate-kustomize.yml
```diff
@@ -0,0 +1,33 @@
...
+          make kustomize
+          echo "$(pwd)/bin" >> $GITHUB_PATH
+
+      - name: Validate all kustomization.yaml files
+        run: |
```
ptoscano, line 24: Ditto.

.github/workflows/validate-kustomize.yml
ptoscano: Let's use `v6`, which is the latest one. Yes, I know, other workflows have older versions; since there is not even consistency among the existing ones (which we ought to update sooner or later), let's use the latest one here.

ptoscano (changes requested):
Thanks for the updates! This mostly LGTM, there are few minor things to fix.

ptoscano:
@coderabbitai resume

ptoscano (approved):
LGTM, thanks!

PR #362, opened 2026-08-07
feat: runtime NetworkPolicies for EDA operator workloads

ttuffin:
Create NetworkPolicy templates and tasks for all EDA operator-managed
workloads: api, ui, default-worker, activation-worker, event-stream,
postgres, and db-management (backup/restore pods).

Pattern follows the gateway operator reference, default-deny first
using match Expressions, per-workload least-privilege allow policies,
conditional managed/external database logic, DNS egress on all
egress-bearing policies, and NP tasks ordered before deployment tasks.

Key EDA-specific decisions:
- Activation-worker has unrestricted egress (user-defined event sources)
- Postgres NP includes app=eda ingress rule for runtime activation Job
  pods that connect via pg_notify (not operator-managed, no managed-by
  label)
- db-management NP covers backup/restore pods blocked by default-deny
- UI NP is conditionally absent when ui_disabled=true

Add networkpolicies to networking.k8s.io RBAC rule and replace molecule
FIXME stub with real NP assertions.

https://redhat.atlassian.net/browse/AAP-73976

## Summary by CodeRabbit

* **New Features**
  * Added network isolation policies for EDA components, including the API, UI, workers, event stream, database management, and activation services.
  * Added controlled PostgreSQL access for managed and external database configurations.
  * Added default-deny ingress and egress protections for EDA-managed workloads.
  * Added conditional UI network policy handling when the UI is disabled.
  * Updated access permissions to support managing Kubernetes NetworkPolicies.
* **Bug Fixes**
  * Improved deployment ordering so database network policies are applied before PostgreSQL pods start.
  * Added validation to confirm expected network policies and access rules are applied.


ttuffin:
@coderabbitai review

molecule/default/tasks/eda_test.yml
```diff
@@ -12,7 +12,62 @@
...
-- name: Add assertions here
-  assert:
-    that: false
-    fail_msg: FIXME Add real assertions for your operator
+- name: Verify default-deny NetworkPolicy exists
+  kubernetes.core.k8s_info:
+    api_version: networking.k8s.io/v1
+    kind: NetworkPolicy
+    namespace: '{{ namespace }}'
+    name: '{{ namespace }}-default-deny'
```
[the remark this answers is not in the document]
ttuffin: Fixed. Added a `set_fact` to derive `eda_cr_name` from the sample CR file at runtime, and updated all three NP lookups to use `eda_cr_name` instead of `namespace`. Also used `supported_pg_version` variable in the postgres NP lookup rather than hardcoding `15`.

molecule/default/tasks/eda_test.yml
```diff
@@ -12,7 +12,62 @@
...
+  until: np_default_deny.resources | length > 0
+  retries: 10
+  delay: 5
+
+- name: Assert default-deny selects eda-operator managed pods and denies all traffic
+  ansible.builtin.assert:
+    that:
+      - np_default_deny.resources[0].spec.podSelector.matchExpressions | length == 1
+      - np_default_deny.resources[0].spec.podSelector.matchExpressions[0].key == 'app.kubernetes.io/managed-by'
+      - "'eda-operator' in np_default_deny.resources[0].spec.podSelector.matchExpressions[0].values"
```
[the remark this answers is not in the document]
ttuffin: Fixed. Added `operator == 'In'` assertion to the default-deny selector check.

molecule/default/tasks/eda_test.yml
```diff
@@ -12,7 +12,62 @@
...
+  until: np_postgres.resources | length > 0
+  retries: 10
+  delay: 5
+
+- name: Assert postgres NetworkPolicy allows app=eda ingress for activation Job pod pg_notify connectivity
+  ansible.builtin.assert:
+    that:
+      - np_postgres.resources[0].spec.podSelector.matchLabels['app.kubernetes.io/component'] == 'database'
+      - np_postgres.resources[0].spec.ingress | length > 0
+      - np_postgres.resources[0].spec.ingress[0].from | selectattr('podSelector.matchLabels.app', 'defined') | selectattr('podSelector.matchLabels.app', 'equalto', 'eda') | list | length > 0
```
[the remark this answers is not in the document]
ttuffin: Declined. The test plan for AAP-73976 (MR !280) explicitly scopes molecule assertions to existence validation and key pod-selector checks, delegating rule-level connectivity validation to ATF regression. Adding per-field port and instance label assertions encodes implementation detail in tests, making them brittle to legitimate rule refinements without adding meaningful coverage. The `app: eda` ingress rule assertion — the one EDA-specific behaviour with real correctness risk — is already validated.

molecule/default/tasks/eda_test.yml
```diff
@@ -12,7 +12,62 @@
...
+  until: np_workload.resources | length > 0
+  retries: 10
+  delay: 5
+  loop:
+    - api
+    - ui
+    - default-worker
+    - activation-worker
+    - event-stream
+    - db-management
```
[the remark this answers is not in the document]
ttuffin: Fixed. The workload loop now conditionally excludes `ui` when `ui_disabled` is true (using a Jinja2 expression on the loop list), and a separate assertion verifies the UI NP is absent in that case. Note: the default CI CR has `ui_disabled: true`, so the loop would have failed on retries for all current CI scenarios.

roles/eda/templates/eda-activation-worker.networkpolicy.yaml.j2
```diff
@@ -0,0 +1,65 @@
...
+              app.kubernetes.io/managed-by: 'aap-gateway-operator'
+      ports:
+        - protocol: TCP
+          port: 8000
+        - protocol: TCP
+          port: 8080
+    # Activation workers create and manage Job pods via the Kubernetes API.
+    # Unrestricted egress is required because event source destinations are
+    # user-defined and cannot be enumerated at policy creation time.
+    - {}
```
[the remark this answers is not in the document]
ttuffin: The explicit granular egress rules have been removed from all four workloads that carry `- {}` (api, activation-worker, default-worker, event-stream). Each now has only DNS egress + `- {}` with a comment explaining why unrestricted egress is required. The `ipBlock` alternative is not viable — event source destinations, external secret manager URLs, and git server hosts are user-configured DNS hostnames, not stable CIDRs, and the operator has no mechanism to resolve them at policy creation time. This pattern is consistent with the gateway operator reference implementation (`gateway.networkpolicy.yaml.j2`).

roles/eda/templates/eda-api.networkpolicy.yaml.j2
```diff
@@ -0,0 +1,85 @@
...
+    # Allow DNS resolution via OpenShift DNS
+    - to:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: openshift-dns
+      ports:
+        - protocol: TCP
+          port: 5353
+        - protocol: UDP
+          port: 5353
```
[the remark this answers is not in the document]
ttuffin: Declined. The EDA server operator targets OpenShift as its primary deployment platform (OCP 4.14+ per SDP requirement R7, ANSTRAT-2322). The gateway operator reference implementation uses the same `openshift-dns:5353` pattern. Adding a `kube-system:53` branch for vanilla Kubernetes is a valid follow-up but is out of scope for this story and would need to be applied consistently across all AAP operators — tracked separately.

roles/eda/templates/eda-api.networkpolicy.yaml.j2
```diff
@@ -0,0 +1,85 @@
...
+              app.kubernetes.io/instance: 'postgres-{{ supported_pg_version }}-{{ ansible_operator_meta.name }}'
+      ports:
+        - protocol: TCP
+          port: {{ eda_postgres_port | default('5432') }}
+{% else %}
+    # External database — destination is user-configured, allow on the port
+    - ports:
+        - protocol: TCP
+          port: {{ eda_postgres_port | default('5432') }}
+{% endif %}
```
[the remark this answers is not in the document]
ttuffin: Declined. The external database host is a user-configured DNS hostname — `ipBlock` cannot match hostnames and the operator has no mechanism to resolve them at policy creation time. A port-only egress rule (no `to`) is the correct approach for external services with unknown destinations, consistent with how the gateway operator reference implementation handles external postgres and redis egress.

roles/eda/templates/eda-ui.networkpolicy.yaml.j2
```diff
@@ -0,0 +1,66 @@
...
+                operator: In
+                values:
+                  - '{{ deployment_type }}-operator'
+                  - 'aap-gateway-operator'
+        - podSelector:
+            matchLabels:
+              control-plane: controller-manager
+      ports:
+        - protocol: TCP
+          port: {{ api_nginx_port }}
```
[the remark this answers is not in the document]
ttuffin: Fixed. Both ingress rules (router and gateway) now use port `8080`, matching the port the UI nginx container actually listens on. The egress rule to the API pod retains `api_nginx_port` (8000). Note: this only affects standalone eda-server-operator deployments — in AAP bundle deployments `ui_disabled` is `true` and this NP is absent entirely.

ttuffin:
Responding to the three nitpick comments from the CodeRabbit review summary:

**`supported_pg_version` in shared vars** — The `roles/backup/vars/main.yml` and `roles/restore/vars/main.yml` in this repo both define `supported_pg_version` independently in their own role scope — this is the established pattern. Moving it to play-level group_vars would be a broader refactor across multiple roles and is out of scope for this story. The fix (adding it to `roles/eda/defaults/main.yml`) follows the same convention already used by the backup and restore roles.

**`apply: true` on the UI absent task** — Fixed. Removed `apply: true` from the UI NP task since it is meaningless when `state: absent`.

**Validate workload policy rules, not only resource existence** — Declined per the test plan for AAP-73976 (MR !280), which explicitly delegates rule-level connectivity validation to ATF regression and scopes molecule to existence validation and key selector checks. The `app: eda` postgres ingress assertion covers the one EDA-specific correctness risk.

molecule/default/tasks/eda_test.yml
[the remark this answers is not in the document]
ttuffin: Fixed. Added `values | length == 1` and replaced the `in` check with an exact equality assertion using `deployment_type + '-operator'`, which derives the expected value the same way the template does rather than hardcoding the string.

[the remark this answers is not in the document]
ttuffin: Fixed.

roles/eda/tasks/networkpolicy.yml
```diff
@@ -0,0 +1,41 @@
...
+
+- name: Apply EDA API NetworkPolicy
+  kubernetes.core.k8s:
+    state: present
+    apply: true
+    definition: "{{ lookup('template', 'eda-api.networkpolicy.yaml.j2') }}"
+
+- name: Apply EDA UI NetworkPolicy
+  kubernetes.core.k8s:
+    state: "{{ 'present' if not ui_disabled else 'absent' }}"
```
kaiokmo, line 20: should we also add an `apply` here for consistency? it doesn't necessarily change anything, and `state` does the job correctly... but still, we're doing it for every other resource, so... I think we should put it here as well:

```suggestion
    state: "{{ 'present' if not ui_disabled else 'absent' }}"
    apply: "{{ omit if ui_disabled else true }}"
```

roles/eda/defaults/main.yml
```diff
@@ -8,6 +8,9 @@ api_version: '{{ deployment_type }}.ansible.com/v1alpha1'
 no_log: true
 ui_disabled: false
 
+# Keep in sync with roles/postgres/vars/main.yml
+supported_pg_version: 15
```
kaiokmo: I think this is OK for now, but we could accidentally drift in case the postgres role changes its value. Why don't we load this value from the postgres role instead? We could `include_vars` before calling calling `Apply PostgreSQL NetworkPolicy`. WDYT?

kaiokmo (approved):

roles/eda/tasks/networkpolicy.yml
ttuffin: I had it originally but removed it after CodeRabbit's [nitpick comment](https://github.com/ansible/eda-server-operator/pull/362#pullrequestreview-4883277148) regarding `state: absent, apply: true` being a meaningless combination.

roles/eda/defaults/main.yml
ttuffin: Makes sense; implemented.

roles/eda/templates/eda-activation-worker.networkpolicy.yaml.j2
```diff
@@ -0,0 +1,40 @@
...
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/name: '{{ ansible_operator_meta.name }}'
+      app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
+      app.kubernetes.io/component: '{{ deployment_type }}-activation-worker'
+  policyTypes:
+    - Ingress
+    - Egress
+  ingress:
```
rooftopcellist: **ANSTRAT-2322 [MAJOR]: Ingress rule missing `ports` field — all ports permitted**

The ingress rule has no `ports` field, allowing all TCP/UDP ports from matched pods. All ingress rules must specify explicit port numbers and protocols.

Determine the port this workload listens on and add it:
```yaml
    ports:
      - protocol: TCP
        port: <actual_port>
```
If this workload accepts no inbound connections from peer pods, remove the `from:` rule and drop `Ingress` from `policyTypes`.

roles/eda/templates/eda-default-worker.networkpolicy.yaml.j2
```diff
@@ -0,0 +1,41 @@
...
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/name: '{{ ansible_operator_meta.name }}'
+      app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
+      app.kubernetes.io/component: '{{ deployment_type }}-default-worker'
+  policyTypes:
+    - Ingress
+    - Egress
+  ingress:
```
rooftopcellist: **ANSTRAT-2322 [MAJOR]: Ingress rule missing `ports` field — all ports permitted**

Same issue as eda-activation-worker. Add an explicit port, or remove the ingress rule and drop `Ingress` from `policyTypes` if this worker accepts no inbound connections.

roles/eda/tasks/networkpolicy.yml
```diff
@@ -0,0 +1,45 @@
...
+    apply: true
+    definition: "{{ lookup('template', 'default-deny.networkpolicy.yaml.j2') }}"
+
+- name: Apply EDA API NetworkPolicy
+  kubernetes.core.k8s:
+    state: present
+    apply: true
+    definition: "{{ lookup('template', 'eda-api.networkpolicy.yaml.j2') }}"
+
+- name: Apply EDA UI NetworkPolicy
```
rooftopcellist, line 18: **ANSTRAT-2322 [MAJOR]: EDA UI NetworkPolicy task missing `apply: true`**

This is the only `kubernetes.core.k8s` NetworkPolicy task in this file without `apply: true`. Without server-side apply, repeated reconciliations risk field-manager conflicts.

Add `apply: true` to match all other NP tasks in the file.

rooftopcellist (changes requested):
ANSTRAT-2322 NetworkPolicy compliance review: two major findings. See inline comments.

roles/eda/tasks/networkpolicy.yml
ttuffin: I had this in originally but CodeRabbit called it out as redundant, so I removed it. Kaio also flagged that it should be here, so I've gone ahead and added it back.

roles/eda/templates/eda-activation-worker.networkpolicy.yaml.j2
```diff
@@ -0,0 +1,40 @@
...
+    # Allow traffic only from other EDA operator-managed pods (task dispatch)
+    - from:
+        - podSelector:
+            matchLabels:
+              app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
```
rooftopcellist: Remove this ingress block entirely. EDA workers use a pull model — the activation-worker connects out to poll for work and never receives inbound connections. The 'task dispatch' comment isn't accurate here: dispatch happens via the worker calling out to the API or Redis, not via inbound pushes to the worker.

Change `policyTypes` to `[Egress]` only and drop the `ingress:` section. This is the same fix that was applied to galaxy-worker in galaxy-operator PR #278.

roles/eda/templates/eda-default-worker.networkpolicy.yaml.j2
```diff
@@ -0,0 +1,41 @@
...
+    # Allow traffic only from other EDA operator-managed pods (e.g. api pod
+    # dispatching tasks to the worker via dispatcherd/postgres LISTEN/NOTIFY)
+    - from:
+        - podSelector:
+            matchLabels:
+              app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
```
rooftopcellist: Remove this ingress block. The postgres LISTEN/NOTIFY mechanism works over the worker's existing outbound connection to postgres — postgres does not open a new inbound connection back to the worker. No port needs to be open on the worker for this to function.

Change `policyTypes` to `[Egress]` only and drop the `ingress:` section. A portless ingress rule allowing all traffic from any operator-managed pod on all ports defeats the purpose of the NetworkPolicy.

roles/eda/templates/db-management.networkpolicy.yaml.j2
```diff
@@ -0,0 +1,42 @@
...
+  namespace: '{{ ansible_operator_meta.namespace }}'
+  labels:
+    {{ lookup("template", "../common/templates/labels/common.yaml.j2") | indent(4) | trim }}
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/managed-by: '{{ deployment_type }}-operator'
+      app.kubernetes.io/component: '{{ deployment_type }}'
+  policyTypes:
+    - Egress
```
rooftopcellist, line 16: Should be `policyTypes: [Ingress, Egress]` with no `ingress:` rules. Declaring only `Egress` leaves ingress governance entirely to the catch-all default-deny NP. If the default-deny is ever absent — maintenance window, namespace migration, disaster recovery — this pod would have no ingress protection at all. Each NP should be self-contained in what it governs.

config/rbac/role.yaml
```diff
@@ -81,13 +81,24 @@ rules:
       - watch
   - apiGroups:
       - apps
```
rooftopcellist: Duplicate `apps` apiGroup rule — there's an identical block earlier in the file. This looks like a leftover from splitting the `networking.k8s.io` rule into its own block. Kubernetes ignores duplicate rules at runtime but it makes the ClusterRole harder to audit. Remove this block.

rooftopcellist (changes requested):
Two prior major findings are still open on the worker NPs, and two new findings surfaced in this round. Details are inline.

roles/eda/templates/eda-activation-worker.networkpolicy.yaml.j2
ttuffin: Removed

roles/eda/templates/eda-default-worker.networkpolicy.yaml.j2
ttuffin: Removed

roles/eda/templates/db-management.networkpolicy.yaml.j2
ttuffin: Added Ingress

config/rbac/role.yaml
ttuffin: Fixed

rooftopcellist (approved):
All open findings addressed. Activation-worker and default-worker are now correctly pure-egress with no ingress rules. db-management policyTypes now covers both Ingress and Egress. Duplicate RBAC rule removed. apply: true is consistent across all NP tasks. Approved.

PR #363, opened 2026-08-25
fix: preserve backup_dir on finalizer path for clean_backup_on_delete

jamesmarshall24:
The postgres_skip_data refactor (#353) moved backup_dir/timestamp
setup into init.yml, which finalizer.yml also runs on CR deletion.
set_fact overwrote the block-scoped backup_dir passed in from CR
status, so delete_backup.yml deleted a freshly created empty
directory instead of the real backup, silently breaking
clean_backup_on_delete.

Guards the three affected tasks with `when: not finalizer_run | bool`,
matching the existing finalizer_run branching in main.yml.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented backup initialization and backup-directory creation during finalizer execution.
  * Ensured backup timestamps and directory naming are handled only during regular runs.

### DIFF
--- a/roles/backup/tasks/init.yml
+++ b/roles/backup/tasks/init.yml
@@ -97,10 +97,12 @@
 - name: Determine the timestamp for the backup once for all nodes
   set_fact:
     now: '{{ lookup("pipe", "date +%F-%T") }}'
+  when: not finalizer_run | bool
 
 - name: Set backup directory name
   set_fact:
     backup_dir: "/backups/eda-openshift-backup-{{ now }}"
+  when: not finalizer_run | bool
 
 - name: Create directory for backup
   k8s_exec:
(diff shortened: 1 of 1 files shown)


ptoscano:
> Guards the three affected tasks with `when: not finalizer_run | bool`, matching the existing finalizer_run branching in main.yml.

Nitpick: maybe moved these 3 tasks in an own `block` guarded by the newly added condition? This way the condition is not repeated, and the tasks belonging to backups are logically grouped.

rooftopcellist (approved):
Same regression and same fix as ansible/awx-operator#2135 — verified the call graph here matches (finalizer.yml sets a block-scoped backup_dir from CR status, init.yml's unconditional set_fact was overwriting it). Guard looks correct. LGTM.

PR #364, opened 2026-08-27
[AAP-89728]fix: mounts the ca bundle correctly and does some DRY work

AlexSCorey:
addresses AAP-89728 and properly mounts the ca-bundle to the eda-api pod.  It also does a little refactoring to dry out some templating that is used by the eda-api and the daphne pods.

## Summary by CodeRabbit

* **New Features**
  * Added optional mounting of the CA trust directory and bundle certificate into containers.
  * Mounted the bundle certificate as read-only when CA bundling is enabled.

* **Refactor**
  * Standardized CA certificate volume mounts across API and Daphne containers for more consistent configuration.

### DIFF
--- a/roles/eda/templates/ca-mount-fragment.yaml.j2
+++ b/roles/eda/templates/ca-mount-fragment.yaml.j2
@@ -1,0 +1,8 @@
+{% if bundle_ca_crt %}
+- name: "ca-trust-extracted"
+  mountPath: "/etc/pki/ca-trust/extracted"
+- name: "{{ ansible_operator_meta.name }}-bundle-cacert"
+  mountPath: /etc/pki/ca-trust/source/anchors/bundle-ca.crt
+  subPath: bundle-ca.crt
+  readOnly: true
+{% endif %}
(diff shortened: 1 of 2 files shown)


kaiokmo (commented):
I'm confused. Why's this needed? I thought the issue of the CA bundle not being attached to the api pods was previously resolved on #342, and also previously reported on AAP-72822.

I understand the refactor itself here, but I don't see how that fixes an issue that was supposed to be already resolved.

On another note, AAP-89728 mentions that `it doesn't attach the CA Bundle to the EDA API Pods`. This is not true. See: https://github.com/ansible/eda-server-operator/blob/cd0d2118845488271a2201cfca7fcaa38c801afb/roles/eda/templates/eda-api.deployment.yaml.j2#L365-L372

AlexSCorey:
> I'm confused. Why's this needed? I thought the issue of the CA bundle not being attached to the api pods was previously resolved on #342, and also previously reported on AAP-72822.
> 
> I understand the refactor itself here, but I don't see how that fixes an issue that was supposed to be already resolved.
> 
> On another note, AAP-89728 mentions that `it doesn't attach the CA Bundle to the EDA API Pods`. This is not true. See:
> 
> https://github.com/ansible/eda-server-operator/blob/cd0d2118845488271a2201cfca7fcaa38c801afb/roles/eda/templates/eda-api.deployment.yaml.j2#L365-L372

@kaiokmo  Apparently it was not being mounted in the api pod. https://redhat.atlassian.net/browse/AAP-89728

AlexSCorey:
This change isn't needed.  Closing.

PR #365, opened 2026-09-16
fix: [AAP-81880] disable snake_case parameter conversion for the EDA kind

snagoor:
## Problem

`topology_spread_constraints` set on an EDA CR never make it into the generated Deployments.
The reconcile fails with an HTTP 422:

```
spec.template.spec.topologySpreadConstraints[0].maxSkew: Invalid value: 0: must be greater than zero
spec.template.spec.topologySpreadConstraints[0].topologyKey: Required value: can not be empty
```

## Cause

The `EDA` entry in `watches.yaml` was missing `snakeCaseParameters: False` (the `EDABackup` and
`EDARestore` entries already have it). Without it, ansible-operator snake_cases the CR spec
*recursively*, so `maxSkew` → `max_skew`, `topologyKey` → `topology_key`, and so on. Those keys get
rendered into the Deployment, the API server prunes them as unknown fields, and the required fields
are left unset.

The Jinja templates are fine — they were already fixed in #275.

## Fix

Add `snakeCaseParameters: False` to the `EDA` watch entry. Every EDA spec field the roles consume is
already snake_case, so the conversion isn't needed; the only camelCase fields in the CRD are native
Kubernetes structures that must stay camelCase.

This also fixes `affinity` (same problem) and `strategy.rollingUpdate`, which was silently falling
back to its `25%` default instead of using the configured values.

## Testing

Added a molecule check that patches the CR with a constraint and asserts the API Deployment renders
it with camelCase keys and the CR's values.

Note: `watches.yaml` is baked into the operator image, so this needs an image rebuild, not just a CR
reapply.

Assisted-by: Claude Opus 5 <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring topology spread constraints for EDA API deployment pods.
  * API pods can now be distributed across nodes according to Kubernetes scheduling settings, including hostname-based spreading and scheduling preferences.

* **Bug Fixes**
  * Preserved Kubernetes-native camelCase fields in topology spread constraints, preventing configuration values from being altered or discarded when applied.


kaiokmo (commented):
Hi @snagoor, can you please rebase your branch on top of main? We had a CI issue that needed fixing. Thanks!

snagoor:
> Hi @snagoor, can you please rebase your branch on top of main? We had a CI issue that needed fixing. Thanks!

Thank you @kaiokmo for the review. Now its rebased and pushed.

ttuffin (approved):
LGTM, thanks for the PR

kaiokmo (changes requested):
@snagoor I have just one small observation: since the main intent of this PR/commit became adding the topology spread constraints feature, could you please amend the commit so it reflects that in the title and make it meaningful for this purpose? The fix is something that needed to be done regardless so it can be explained in the commit body, but the main contribution here is the topology spread constraints, so I think the title could be updated.

Thanks in advance!

snagoor:
> @snagoor I have just one small observation: since the main intent of this PR/commit became adding the topology spread constraints feature, could you please amend the commit so it reflects that in the title and make it meaningful for this purpose? The fix is something that needed to be done regardless so it can be explained in the commit body, but the main contribution here is the topology spread constraints, so I think the title could be updated.
> 
> Thanks in advance!

Updated as requested.

kaiokmo (approved):
thank you!

PR #366, opened 2026-09-16
fix: [AAP-91811] replace is_defined guards with default('') for VSO secret support

sunalawa:
event_stream.database_secret and event_persistence.database_secret used
'is defined' guards which break when an empty-string default is added —
the same pattern fixed in galaxy-operator [PR #283](https://github.com/ansible/galaxy-operator/pull/283). Replaced with
'| default('') | length' so both undefined and '' are treated as unset,
allowing VSO-pre-created secrets to be correctly preserved on reconcile.

Ref: AAP-91811, ANSTRAT-2212
Signed-off-by: Suyash Nalawade <sunalawa@redhat.com>

## Summary by CodeRabbit

* **Refactor**
  * Simplified PostgreSQL configuration checks while preserving existing behavior.
  * Secret validation continues to run only when non-empty values are provided.
  * Tasks remain skipped when relevant secrets are missing or empty.


molecule/default/tasks/postgres_secret_guard_test.yml
```diff
@@ -0,0 +1,154 @@
+---
+- name: Assert guard test variables are defined
+  ansible.builtin.assert:
+    that:
+      - namespace is defined and namespace | length > 0
+      - guard_test_event_stream_secret is defined and guard_test_event_stream_secret | length > 0
+      - guard_test_sentinel is defined and guard_test_sentinel | length > 0
+      - guard_test_event_persistence_generated_suffix is defined
+      - guard_test_event_stream_generated_suffix is defined
+    fail_msg: "Required AAP-91811 guard test variables are not defined (check molecule.yml group_vars)"
```
kaiokmo: Nit: given this is a molecule test, I don't think we need to mention the Jira ticket number here. 
```suggestion
    fail_msg: "Required guard test variables are not defined (check molecule.yml group_vars)"
```

molecule/default/tasks/postgres_secret_guard_test.yml
```diff
@@ -0,0 +1,154 @@
...
+  delay: 15
+
+- name: Assert both database configuration secret status fields are defined
+  ansible.builtin.assert:
+    that:
+      - eda_cr.resources[0].status.eventStreamDatabaseConfigurationSecret is defined
+      - eda_cr.resources[0].status.eventPersistenceDatabaseConfigurationSecret is defined
+    fail_msg: >-
+      Operator did not publish both DB configuration secret status fields.
+      status={{ eda_cr.resources[0].status | default({}) }}
```
kaiokmo: I feel that this test is redundant and can be removed. The `until` in the previous "Read the EDA CR status" task already serves the purpose of checking that both fields are present. So, if the previous test passes, both of the `is defined` here are necessarily true. Therefore, they become redundant.
```suggestion
```

molecule/default/tasks/postgres_secret_guard_test.yml
```diff
@@ -0,0 +1,154 @@
...
+
+- name: Assert the boolean guard expression evaluates correctly
+  vars:
+    empty_component: {}
+  ansible.builtin.assert:
+    that:
+      - (guard_test_event_stream_secret | default('') | length > 0) | bool
+      - not ('' | default('') | length > 0) | bool
+      - not (empty_component.database_secret | default('') | length > 0) | bool
+    fail_msg: "The `| length > 0` boolean guard cast did not evaluate as expected"
```
kaiokmo: I think that these are not adding much value to operator testing. Overall, they're testing Jinja/Ansible semantics, not the operator itself. 
```suggestion
```

molecule/default/tasks/postgres_secret_guard_test.yml
```diff
@@ -0,0 +1,154 @@
...
+            name: generated-default-secret
+  ansible.builtin.assert:
+    that:
+      - (custom_found['resources'] | default([]) | length
+         | ternary(custom_found, default_cfg))['resources'][0]['metadata']['name'] == 'adopted-external-secret'
+      - (custom_absent['resources'] | default([]) | length
+         | ternary(custom_absent, default_cfg))['resources'][0]['metadata']['name'] == 'generated-default-secret'
+      - (not custom_absent['resources'] | default([]) | length) | bool
+      - not ((not custom_found['resources'] | default([]) | length) | bool)
+    fail_msg: "The adopt-vs-generate ternary guard logic did not select the expected configuration"
```
kaiokmo: I also think these are just testing Jinja mechanics.
```suggestion
```

kaiokmo (changes requested):

molecule/default/tasks/postgres_secret_guard_test.yml
sunalawa: Addressed the fix by removing the redundant test in d60d68f50b1898bea137e9dd2ddf1ece87cd13c8

molecule/default/tasks/postgres_secret_guard_test.yml
sunalawa: Addressed the fix by removing the redundant test in d60d68f50b1898bea137e9dd2ddf1ece87cd13c8

molecule/default/tasks/postgres_secret_guard_test.yml
sunalawa: Addressed the fix by removing the redundant test in d60d68f50b1898bea137e9dd2ddf1ece87cd13c8

molecule/default/tasks/postgres_secret_guard_test.yml
sunalawa: Addressed the fix by removing the redundant test in d60d68f50b1898bea137e9dd2ddf1ece87cd13c8

kaiokmo (approved):

